PR #58, opened 2024-08-05
feat: print DTS generation time

Timeless0911:
## Summary

print DTS generation time

## Checklist

### DIFF
--- a/packages/plugin-dts/src/dts.ts
+++ b/packages/plugin-dts/src/dts.ts
@@ -6,7 +6,7 @@ import { emitDts } from './tsc';
 import { ensureTempDeclarationDir, loadTsconfig } from './utils';
 
 export async function generateDts(data: DtsGenOptions) {
-  logger.info('Generating DTS...');
+  logger.ready('Generating DTS...');
   const { options: pluginOptions, cwd, isWatch } = data;
   const { tsconfigPath, distPath, bundle, entryPath } = pluginOptions;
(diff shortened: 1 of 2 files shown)


packages/plugin-dts/src/dts.ts
```diff
@@ -6,7 +6,7 @@ import { emitDts } from './tsc';
 import { ensureTempDeclarationDir, loadTsconfig } from './utils';
 
 export async function generateDts(data: DtsGenOptions) {
-  logger.info('Generating DTS...');
+  logger.ready('Generating DTS...');
```
chenjiahan, line 9: Is `start` more suitable here?

packages/plugin-dts/src/tsc.ts
```diff
@@ -58,10 +62,10 @@ export function emitDts(
         logger.error(message);
       }
 
-      throw new Error('TypeScript compilation failed');
+      throw new Error('DTS generation failed');
```
fi3ework, line 65: User `.d.ts.` instead of `DTS` in user side?

Timeless0911: How about using `declarations files` instead of `DTS`?

Timeless0911: we can add specific suffix file names after we support `autoExtensions` in DTS

fi3ework: "DTS" does not seem to be a formal term. We can change it in the future based on how the log looks in the end.

fi3ework (approved):

PR #59, opened 2024-08-05
chore(infra/e2e): @e2e/helper rather than #imports

SoonIter:
## Summary

at present, e2e cases with `package.json` ignore the imports field and it is too inconvenient

https://github.com/web-infra-dev/rslib/blob/864c46eb3fd6edbc336e926a7a92aeb7511c2f99/e2e/cases/autoExtension/type-commonjs/rslib.config.ts#L5

so, we can align with rsbuild, use `@e2e/helper` package to solve this issue


fi3ework:
Glad to see one more alignment.

fi3ework (approved):

PR #65, opened 2024-08-06
test: add tests of dts: false and dts.distPath

Timeless0911:
## Summary

add tests of dts: false and dts.distPath

## Checklist

### DIFF
--- a/e2e/cases/dts/bundle/rslib.false.config.ts
+++ b/e2e/cases/dts/bundle/rslib.false.config.ts
@@ -1,0 +1,16 @@
+import { generateBundleCjsConfig, generateBundleEsmConfig } from '@e2e/helper';
+import { defineConfig } from '@rslib/core';
+
+export default defineConfig({
+  lib: [
+    generateBundleEsmConfig(__dirname, {
+      dts: false,
+    }),
+    generateBundleCjsConfig(__dirname),
+  ],
+  source: {
+    entry: {
+      main: './src/index.ts',
+    },
+  },
+});
--- a/e2e/cases/dts/index.test.ts
+++ b/e2e/cases/dts/index.test.ts
@@ -1,20 +1,76 @@
 import { join } from 'node:path';
 import { buildAndGetResults } from '@e2e/helper';
-import { expect, test } from 'vitest';
+import { describe, expect, test } from 'vitest';
 
-test('dts when bundle: false', async () => {
-  const fixturePath = join(__dirname, 'bundle-false');
-  const { files, contents } = await buildAndGetResults(fixturePath, 'dts');
+describe('dts when bundle: false', () => {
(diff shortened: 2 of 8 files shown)


e2e/cases/dts/index.test.ts
```diff
@@ -1,20 +1,76 @@
 import { join } from 'node:path';
 import { buildAndGetResults } from '@e2e/helper';
-import { expect, test } from 'vitest';
+import { describe, expect, test } from 'vitest';
 
-test('dts when bundle: false', async () => {
-  const fixturePath = join(__dirname, 'bundle-false');
-  const { files, contents } = await buildAndGetResults(fixturePath, 'dts');
+describe('dts when bundle: false', () => {
+  test('basic - bundleless dts', async () => {
```
fi3ework, line 6: We could use nested `describe` instead of manual `basic -` prefix, and it will be easier to test with filter.

e2e/cases/dts/bundle/rslib.false.config.ts
```diff
@@ -0,0 +1,16 @@
+import { generateBundleCjsConfig, generateBundleEsmConfig } from '@e2e/helper';
+import { defineConfig } from '@rslib/core';
+
+export default defineConfig({
+  lib: [
+    generateBundleEsmConfig(__dirname, {
```
fi3ework, line 6: The name is too long and verbose, maybe we could change it to `config.esm()` & `config.cjs()` int the future.

e2e/cases/dts/index.test.ts
```diff
@@ -1,20 +1,76 @@
...
+
+  test('distPath - bundleless dts', async () => {
+    const fixturePath = join(__dirname, 'bundle-false');
+    const { files } = await buildAndGetResults(
+      fixturePath,
+      'rslib.distpath.config.ts',
+      'dts',
+    );
+    expect(files.esm?.length).toBe(4);
+    expect(files.esm?.[0]!.startsWith('./dist/custom')).toEqual(true);
```
fi3ework, line 38: Is the order in files stable?

Timeless0911: The order doesn't matter, we only need to check any file whether it located in correct distpath or not.

e2e/cases/dts/bundle/rslib.false.config.ts
Timeless0911: In the future, we will definitely do a test related refactoring.

e2e/cases/dts/index.test.ts
Timeless0911: Snapshot cannot be used outside of test, nested describe is not a good choice. I think we can simply remove `- xxx`

image
 
The log is clear enough

PR #69, opened 2024-08-07
feat: support dts.abortOnError

Timeless0911:
## Summary

support `dts.abortOnError` option to control whether to exit process when dts build error.

## Checklist

### DIFF
--- a/packages/plugin-dts/src/index.ts
+++ b/packages/plugin-dts/src/index.ts
@@ -1,10 +1,11 @@
 import { fork } from 'node:child_process';
 import { extname, join } from 'node:path';
-import type { RsbuildPlugin } from '@rsbuild/core';
+import { type RsbuildPlugin, logger } from '@rsbuild/core';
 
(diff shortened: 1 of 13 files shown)


packages/plugin-dts/src/index.ts
```diff
@@ -79,7 +79,28 @@ export const pluginDts = (options: PluginDtsOptions): RsbuildPlugin => ({
...
+    });
+
+    api.onAfterBuild({
+      handler: ({ isFirstCompile }) => {
+        if (!isFirstCompile) {
+          return;
+        }
+
+        for (const result of promisesResult) {
+          if (result.startsWith('Error')) {
```
fi3ework, line 92: I know using a `string` works here. But what about using a structured object, e.g. `{ status: 'error' | 'success' }`, which is way easier to extend in the future.

Timeless0911: good idea

fi3ework (approved):

PR #71, opened 2024-08-07
fix: lcp should not be a file path

### LINKED ISSUE #68: [Bug]: Single file in bundleless mode outputs error
### Details

Currently only have one file `src/index.ts`, and generate bundleless outputs

```ts
source: {
  entry: {
    main: ['./src/**'],
  },
},
```

It will behave like 
image

Entry calculation should deal with single file situation.

fi3ework:
## Summary

Fix #68.

Changes overview:

1. fix #68, don't return file path as a LCP.
2. add fs mock by following https://vitest.dev/guide/mocking.html#file-system to facilitate testing with fs/path.
3. add separated vitest config to make sure the `packages/core/__mocks__` could be resovled, or we have to put it in the project root :(.

## Checklist


e2e/cases/bundle-false/index.test.ts
```diff
@@ -2,12 +2,40 @@ import { join } from 'node:path';
...
-test('bundle: false', async () => {
+test('basic', async () => {
   const fixturePath = join(__dirname, 'basic');
   const { files } = await buildAndGetResults(fixturePath);
 
-  // TODO: record file paths with inline snapshot
-  // need to add path serialization
-  expect(files.esm?.length).toBe(4);
-  expect(files.cjs?.length).toBe(4);
+  expect(files.esm?.sort()).toMatchInlineSnapshot(`
```
Timeless0911, line 9: We can sort in `buildAndGetResults`?

packages/core/src/utils/extension.ts
```diff
@@ -32,7 +43,9 @@ export const getDefaultExtension = (options: {
     );
```
Timeless0911, line 43: use `fs.readFileSync(pkgJsonPath, 'utf8'),`

Timeless0911 (approved):

PR #72, opened 2024-08-08
feat: support DTS autoExtension

Timeless0911:
## Summary

Support DTS autoExtension
- `.d.ts` with esm format and `.d.cts` with cjs format when `type: module`
- `.d.ts` with cjs format and `.d.mts` with esm format when `type: commonjs`

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -168,19 +168,25 @@ const composeAutoExtensionConfig = (
   format: Format,
   root: string,
   autoExtension: boolean,
-): RsbuildConfig => {
-  const { jsExtension } = getDefaultExtension({
+): {
+  config: RsbuildConfig;
+  dtsExtension: string;
+} => {
+  const { jsExtension, dtsExtension } = getDefaultExtension({
     format,
     root,
     autoExtension,
   });
 
   return {
-    output: {
-      filename: {
-        js: `[name]${jsExtension}`,
+    config: {
+      output: {
+        filename: {
+          js: `[name]${jsExtension}`,
+        },
       },
     },
+    dtsExtension,
   };
 };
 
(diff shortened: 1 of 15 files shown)


packages/core/src/config.ts
```diff
@@ -405,7 +410,7 @@ async function composeLibRsbuildConfig(
     config.bundle,
     dirname(configPath),
   );
-  const dtsConfig = await composeDtsConfig(config);
+  const dtsConfig = await composeDtsConfig(config, dtsExtension);
```
fi3ework, line 413: Could we move the dts logic in `composeAutoExtensionConfig` into `composeDtsConfig`, then DTS config will not couple with JS anymore.

packages/plugin-dts/src/apiExtractor.ts
```diff
@@ -5,23 +5,34 @@ import {
...
     entry = 'index.d.ts',
     tsconfigPath = 'tsconfig.json',
   } = options;
   try {
-    const untrimmedFilePath = join(cwd, relative(cwd, outDir), 'index.d.ts');
+    const start = Date.now();
+    const untrimmedFilePath = join(
+      cwd,
+      relative(cwd, outDir),
+      `index${dtsExtension}`,
```
fi3ework, line 34: The `index` is fixed for now?

Timeless0911: What should our bundled DTS file named

packages/core/src/config.ts
Timeless0911: I aim to reduce multiple time call for `getDefaultExtension`
image
 Or is this acceptable?

packages/plugin-dts/src/apiExtractor.ts
fi3ework: Provide naming config to user, we could implement this in the future.

packages/core/src/config.ts
fi3ework: So be it like this, we could do some refactor in the future.

fi3ework (approved):

PR #74, opened 2024-08-08
feat(bundle): support autoExternal configuration

9aoy:
## Summary

support automatically externalize project dependencies and peerDependencies and not package them into the final bundle.

```
type AutoExternal =
  | boolean
  | {
      dependencies?: boolean;
      devDependencies?: boolean;
      peerDependencies?: boolean;
    };
```

## Checklist


packages/core/src/config.ts
```diff
@@ -394,17 +399,24 @@ async function composeLibRsbuildConfig(
...
   const { config: autoExtensionConfig, dtsExtension } =
-    composeAutoExtensionConfig(format!, dirname(configPath), autoExtension);
+    composeAutoExtensionConfig(format!, autoExtension, pkgJson);
   const bundleConfig = composeBundleConfig(config.bundle);
   const targetConfig = composeTargetConfig(config.output?.target);
   const syntaxConfig = composeSyntaxConfig(
     config.output?.syntax,
     config.output?.target,
   );
+  const autoExternalConfig = composeAutoExternal({
```
Timeless0911, line 415: `composeAutoExternalConfig`?

Timeless0911:
Can we update examples `externals` config?

packages/core/src/utils/external.ts
```diff
@@ -0,0 +1,58 @@
+import type { RsbuildConfig } from '@rsbuild/core';
+import type { PkgJson } from '../types';
+import type { AutoExternal } from '../types/config';
+import { logger } from './logger';
+
+export const composeAutoExternalConfig = (options: {
```
fi3ework, line 6: Why not put it in `config.ts`?

packages/core/src/utils/external.ts
```diff
@@ -0,0 +1,58 @@
...
+    devDependencies: false,
+    ...(autoExternal === true ? {} : autoExternal),
+  };
+
+  // User externals configuration has higher priority than autoExternal
+  // eg: autoExternal: ['react'], user: output: { externals: { react: 'react-1' } }
+  // Only handle the case where the externals type is object, string / string[] does not need to be processed, other types are too complex.
+  const userExternalKeys =
+    userExternals &&
+    Object.prototype.toString.call(userExternals) === '[object Object]'
```
fi3ework, line 36: Does `userExternals instanceof Object` meets the requirement.

packages/core/src/utils/external.ts
```diff
@@ -0,0 +1,58 @@
...
+      ? Object.keys(userExternals)
+      : [];
+
+  const externals = (
+    ['dependencies', 'peerDependencies', 'devDependencies'] as const
+  )
+    .reduce<string[]>((prev, type) => {
+      if (externalOptions[type]) {
+        return pkgJson[type] ? prev.concat(Object.keys(pkgJson[type]!)) : prev;
```
fi3ework, line 45: 1. No deduplication in `prev`.
2. If a dep is in `devDependencies`, is will be externalized as well?

Timeless0911: `if (externalOptions[type])` do judgement?

packages/core/src/utils/external.ts
9aoy:
```
>  /.a/ instanceof Object => true
>  Object.prototype.toString.call(/.a/) => '[object RegExp]'
```

fi3ework: Awesome JaVa$cRipT.

packages/core/src/utils/external.ts
9aoy: `devDependencies` will not externalized by default,  so do we need to support `development` configuration item in autoExternal?

```diff
type AutoExternal =
  | boolean
  | {
      dependencies?: boolean;
-     devDependencies?: boolean;
      peerDependencies?: boolean;
    };
```

fi3ework: I saw that, LGTM, no more 1. question.
If a dep is in `peerDependencies` and `dependencies` (I don't know why user does so), we still need a dedup thought.

9aoy:
> Can we update examples `externals` config?

The example scenario is not yet supported. When we external `react`, `react/jsx-runtime` will not be externalized.  And I think `autoExternal` should support this feature (next PR).

packages/core/src/utils/external.ts
9aoy: Okay, let's dedup. In fact, duplicate keys have no impact on the function. 😆

9aoy: done

chenjiahan:
> When we external react, react/jsx-runtime will not be externalized.

Why is `react/jsx-runtime` not externalized, I thought it should work.

Timeless0911:
https://github.com/web-infra-dev/modern.js/blob/3824a0e819d2444fb220aa96491c2971c998d514/packages/solutions/module-tools/src/config/merge.ts#L91-L105

What Modern.js Module do

9aoy:
> Why is react/jsx-runtime not externalized, I thought it should work.

done

Timeless0911 (approved):

PR #75, opened 2024-08-08
feat: support DTS bundle file name

Timeless0911:
## Summary

support generate bundle DTS name according to entry name.

image

## Checklist

### DIFF
--- a/packages/plugin-dts/src/index.ts
+++ b/packages/plugin-dts/src/index.ts
@@ -60,15 +60,12 @@ export const pluginDts = (options: PluginDtsOptions): RsbuildPlugin => ({
           stdio: 'inherit',
         });
 
-        let dtsEntry: DtsEntry = {
-          name: undefined,
-          path: undefined,
-        };
-
-        if (options.bundle) {
-          // TODO: temporarily use first element of Record<string, string>
-          dtsEntry = processSourceEntry(config.source?.entry);
-        }
+        // TODO: @microsoft/api-extractor only support single entry to bundle DTS
+        // use first element of Record<string, string> type entry config
+        const dtsEntry = processSourceEntry(
+          options.bundle!,
+          config.source?.entry,
+        );
 
         const dtsGenOptions: DtsGenOptions = {
           ...options,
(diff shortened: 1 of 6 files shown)


packages/plugin-dts/src/utils.ts
```diff
@@ -70,3 +71,26 @@ export async function processDtsFiles(
...
+      name: Object.keys(entryConfig)[0] as string,
+      path: Object.values(entryConfig)[0] as string,
+    };
+  }
+
+  logger.warn(
+    '@microsoft/api-extractor only support single entry of Record<string, string> type to bundle DTS.',
+  );
+
+  return {
```
fi3ework, line 92: Is dts will be emitted if return from here?

packages/plugin-dts/src/index.ts
```diff
@@ -54,10 +60,19 @@ export const pluginDts = (options: PluginDtsOptions): RsbuildPlugin => ({
           stdio: 'inherit',
         });
 
+        let dtsEntry: DtsEntry = {
+          name: undefined,
+          path: undefined,
+        };
```
fi3ework, line 66: `processSourceEntry` could directly return a `dtsEntry`, instead another mut processing.

packages/plugin-dts/src/utils.ts
Timeless0911: The bundless DTS will generated and api-extractor will throw error
image

fi3ework: 1. bundleless output is useless as user want the bundled, can we detect the invalid option earlier and do not call the bundleless emit.
2. the log above (in line 88) should turn to be an error.

Timeless0911: I do a judgement and log like this
image

fi3ework: Could put the warn right above the error?

Timeless0911: image
is this OK?

fi3ework (approved):

PR #78, opened 2024-08-09
fix: rootDir in bundle DTS

Timeless0911:
## Summary

no need to specify `rootDir` in `tsconfig.json`

## Checklist

### DIFF
--- a/packages/plugin-dts/src/dts.ts
+++ b/packages/plugin-dts/src/dts.ts
@@ -116,7 +116,8 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
     throw new Error();
   }
   const { options: rawCompilerOptions } = loadTsconfig(configPath);
-  const rootDir = rawCompilerOptions.rootDir ?? 'src';
+  const rootDir =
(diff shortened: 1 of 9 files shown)


packages/plugin-dts/src/dts.ts
```diff
@@ -24,7 +24,8 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
     throw new Error();
   }
   const { options: rawCompilerOptions } = loadTsconfig(configPath);
-  const rootDir = rawCompilerOptions.rootDir ?? 'src';
+  const rootDir =
+    rawCompilerOptions.rootDir ?? join(dirname(configPath), 'src');
```
fi3ework, line 28: Why do we need to add to provide a fallback to `src`, although is commonly used. Is it better to require user to define it explicitly?

Timeless0911: We need rootDir to calculate dts entry file in bundle mode, if user do not config, fallback will help.

fi3ework: In 95% cases, `src` helps. What about in other `5%`.

PR #80, opened 2024-08-12
feat: support autoExternal for dts files

9aoy:
## Summary

Rslib will determine whether third-party types need to be bundled based on autoExternal & externals configurations. (Third-party package types used in devDeps are bundled by default.)

## Checklist


e2e/cases/auto-external/default/rslib.config.ts
```diff
@@ -2,10 +2,23 @@ import { generateBundleCjsConfig, generateBundleEsmConfig } from '@e2e/helper';
 import { defineConfig } from '@rslib/core';
 
 export default defineConfig({
-  lib: [generateBundleEsmConfig(__dirname), generateBundleCjsConfig(__dirname)],
+  lib: [
+    {
+      ...generateBundleEsmConfig(__dirname),
```
Timeless0911, line 7: this function have two params, you don't need to use `...` to assign object.

e2e/scripts/shared.ts
```diff
@@ -132,3 +134,40 @@ export const buildAndGetResults = async (
     isSuccess: Boolean(rsbuildInstance),
   };
 };
+
+export const buildAndGetAllResults = async (
```
Timeless0911, line 138: Can we combine this function with `buildAndGetResults` as `type: all`to reduce repeated code?

Timeless0911 (approved):

PR #86, opened 2024-08-13
chore: init release workflow

Timeless0911:
## Summary

init basic release workflow

## Checklist


.changeset/config.json
```diff
@@ -0,0 +1,15 @@
+{
```
fi3ework, line 1: We use changeset only to bump packages?

.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,48 @@
+name: Release Nightly
```
fi3ework, line 1: If this is copied from Rsbuild repository, we need to ensure nightly workflow is still being used. As for https://github.com/web-infra-dev/rsbuild/actions/workflows/release-nightly.yml, the last it run was 9 months ago.

.changeset/config.json
Timeless0911: yes, we only need to add a changeset before publish instead of in developing PR

.github/workflows/release-nightly.yml
Timeless0911: nightly seems no need now, rsbuild doesnt publish nightly for a long time, maybe we can delete it.

fi3ework: Yeah, canary / alpha(beta) and latest is enough for us.

.changeset/config.json
fi3ework: https://nx.dev/nx-api/nx/documents/release#dryrun seems promising, we could replace changeset in Rsbuild as well. IIRC, cc @chenjiahan wanted to replace the changeset before.

Timeless0911: The `web-infra-dev/actions@v2` does not support nx workflow, maybe we can use a lightweight bump and publish model write by ourselves👀

Timeless0911: Adding changeset by maintainer manually in our daily PR.
And bumping versions are operated in workflow

fi3ework: But Rsbuild had stop adding changeset daily for a time being, and using changesets as a classic Lerna.

Timeless0911: Now they are publishing beta version and they just bump in local and make a PR now. Since former latest version, 
image
they add a changeset before publish or in daily development, just what I do in Rspress now.

fi3ework: web-infra-dev/actions requires changesets now. We may look for the alternatives in the future.

fi3ework (approved):

Issue #46, opened 2024-07-26
Rslib features track


Timeless0911:
# CLI

- [x] Support `npm create rslib@latest` to create a Rslib project
- [x] Support `--inspect` to show Rslib/Rsbuild/Rspack config
- [x] #138
- [x] Add some debug info to better debugging Rslib

# Config

- [x] Provide preset templates to default set some common configs for basic scenes
- [x] Support better shared Rsbuild config

# Features
> We try to keep the features aligned with [Modern.js Module](https://modernjs.dev/module-tools/en/api/index.html)

## bundle / bundleless

- [x] basic support

## format
[output.library.type](https://www.rspack.dev/config/output#outputlibrarytype)
- [x] cjs
- [x] esm
- [x] umd
- [x] mf

## platform
- [x] support [output.target](https://rsbuild.dev/config/output/target#outputtarget)
- [x] support control the syntax target for downgration

## alias
- [x] use [source.alias](https://rsbuild.dev/config/source/alias#sourcealias)

## autoExtension
- [x] js extension
  - [x] `.js` with esm format and `.cjs` with cjs format when `type: module`
  - [x] `.js` with cjs format and `.mjs` with esm format when `type: commonjs`
- [x] dts extension
  - [x] `.d.ts` with esm format and `.d.cts` with cjs format when `type: module`
  - [x] `.d.ts` with cjs format and `.d.mts` with esm format when `type: commonjs`

## externals
[output.externals](https://rsbuild.dev/config/output/externals#outputexternals)
- [x]  prevent externalizing entry modules in bundleless mode
- [x]  make node built-in modules external when target is node
- [x]  support  auto external dependencies and peerDependencies
- [x]  support [tools.swc](https://rsbuild.dev/config/tools/swc) to external helpers  

## redirect
- [x]  redirect in js outputs of bundleless mode
	- [x] path
	- [x] extension (import/export)
	- [x] style
	- [x] asset path
- [x]  redirect in DTS outputs of bundleless mode
	- [x] path
	- [x] extension

## shims
- [x]  shims `__filename` and `__dirname` in esm
- [x]  shims `import.meta.url` in cjs
- [x]  set `module.parser.javascript.importMeta = false` in esm

## banner & footer
- [x]  integrate [banner-plugin](https://www.rspack.dev/plugins/webpack/banner-plugin) to inject content
- [x]  support inject in dts files

## dts
- [x] support `cleanDistPath` to clean DTS output before emit
[...]

PR #87, opened 2024-08-13
chore(cli): add --clean option to build command

shulaoda:
## Related Links
Related to #46 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ This repo has adopted the ByteDance Open Source Code of Conduct. Please check [C
 
 Come and chat with us on [Discord](https://discord.gg/FQfm7VqU)! The Rstack team and users are active there, and we're always looking for contributions.
 
+## 🌟 Quality
+
+Rslib uses [Web Infra QoS](https://web-infra-qos.netlify.app?product=rslib&metrics=bundle-size) to observe the trend of key metrics, such as bundle size, compile speed and install size.
+
 ## 🙏 Credits
 
 Some of the implementations of Rslib have drawn inspiration from outstanding projects in the community. We would like to express our gratitude to them:


packages/core/src/cli/commands.ts
```diff
@@ -41,10 +45,21 @@ export function runCli(): void {
...
   buildCommand
     .option('-w --watch', 'turn on watch mode, watch for changes and rebuild')
+    .option('--clean', 'clean the output directory before building')
     .description('build the library for production')
     .action(async (options: BuildOptions) => {
       try {
         const rslibConfig = await loadConfig(options.config, options.envMode);
+
+        if (options.clean) {
+          const distDir = rslibConfig.output?.distPath?.root ?? OUTPUT_DIST_DIR;
```
Timeless0911, line 55: It's raw Rsbuild config, we should get `distDir` from final config.

Timeless0911: Rsbuild already support clean [output.cleanDistPath](https://rsbuild.dev/zh/config/output/clean-dist-path#outputcleandistpath), may be you should implemeny clean feature in plugin-dts, to add an option of clean, and in default rslib core logic, when we use builtin plugin-dts, pass `clean: true` to it.

Timeless0911: The issue was written early and may have some problems now.

Timeless0911:
I have changed the feature issue 
image
 
What do you think about the clean feature?

packages/core/src/cli/commands.ts
shulaoda: Would it be confusing to include the `clean` option in `plugin-dts`?

shulaoda: Because `clean` is a global build option, not `plugin-dts`.

Timeless0911: We hope `rsbuild-plugin-dts` can be used seprately in a Rsbuild project, maybe we can inherit `output.cleanDistPath` options of Rsbuild and use its value to do DTS output clean?

shulaoda: Maybe I misunderstood. Do we need to clear `DTS` output instead of `dist` output before building?

Timeless0911: Js output will be cleaned according to Rsbuild `output.cleanDistPath` by default, the DTS output should align with this behaviour. Maybe we don't need a cli option to do that, just in built-in logic.

shulaoda: > Maybe we don't need a cli option to do that, just in built-in logic.

Agree

shulaoda: However, `Rsbuild` seems to default to clear `dist`. What is the significance of doing so? Does `plugin-dts` not support default clear?

Timeless0911: The `dist` maybe dirty in local development, clean behavihour can ensure the new output will not be influenced by existed output.

Timeless0911: `plugin-dts` have not supported it yet.

We should clean:
- bundleless mode: the tsc emit path
- bundle mode: the tsc emit temp path `.rslib/declarations` and api-extractor emit path

shulaoda:
I tried to rewrite, but it seems that an empty commit caused PR to close.

fi3ework:
Feel free to re-create one and link to this PR.

Issue #106, opened 2024-08-16
[Bug]: `output.filename.js` is not working even when `autoExtension` is `false`


xc2:
### Version

```sh
System:
    OS: macOS 15.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 53.33 MB / 32.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Browsers:
    Chrome: 127.0.6533.120
  npmPackages:
    @rslib/core: 0.0.2 => 0.0.2
```

### Details

```ts
import { defineConfig } from "@rslib/core";

export default defineConfig({
  source: {
    entry: {
      index: './src/index.ts',
    },
  },
  output: {
    distPath: {
      root: './dist'
    }
  },
  lib: [
    {
      format: "cjs",
      dts: { bundle: false },
      autoExtension: false,
      output: {
        filename: {
          js: '[name].cjs'
        }
      }
    },
    {
      format: "esm",
      dts: false,
    },
  ],
});

```

### Reproduce link

-

### Reproduce Steps

expected `dist/index.cjs` but got `dist/index.js`

Timeless0911:
Got it, I will fix it.

PR #108, opened 2024-08-19
fix: should respect `output.filename.js` config

Timeless0911:
## Summary

closes: #106 

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -240,7 +240,7 @@ const composeFormatConfig = (format: Format): RsbuildConfig => {
 };
 
 const composeAutoExtensionConfig = (
-  format: Format,
+  config: LibConfig,
   autoExtension: boolean,
   pkgJson?: PkgJson,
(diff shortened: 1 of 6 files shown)


packages/core/src/config.ts
```diff
@@ -259,6 +259,7 @@ const composeAutoExtensionConfig = (
       output: {
         filename: {
           js: `[name]${jsExtension}`,
+          ...config.output?.filename,
```
fi3ework, line 262: What about early return if `autoExtension` is disabled, and the `config.output?.filename` won't be overridden.

Timeless0911: Maybe the commit message is not accurate, users can modify `config.output?.filename` no matter `autoExtension` enabled or not. They can customize both filename and extname.

Timeless0911: I will add more tests to cover it.

fi3ework (approved):

PR #124, opened 2024-08-26
feat: add warn when request type bail from "module-import"

fi3ework:
## Summary

When using `require` to load an external module in ESM output, the `require` statement will adopt the `module` external type (the behavior of `"module-import"` external type). However, it can be challenging to identify the desired external type for the `require` statement. As a result, a warning will prompt the user to explicitly specify the correct external type.

Code 2024-08-26 12 54 20

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -166,8 +166,8 @@ export async function createInternalRsbuildConfig(): Promise<RsbuildConfig> {
       },
     },
     output: {
-      filenameHash: false,
       // TODO: easy to development at the moment
+      filenameHash: false,
       minify: false,
       distPath: {
         js: './',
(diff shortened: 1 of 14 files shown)


packages/core/src/config.ts
```diff
@@ -239,6 +236,71 @@ const composeFormatConfig = (format: Format): RsbuildConfig => {
   }
 };
 
+export const composeModuleImportWarn = (request: string): string => {
+  return `The externalized commonjs request "${request}" will use "module" external type in ESM format. If you want to specify other external type, considering set "output.externals" with the request and type.`;
```
Timeless0911, line 240: how about use a purple or other color for key words like `${request}` `"module"` `"output.externals"`

fi3ework: Good idea, I'll give a try.

fi3ework: image

Timeless0911 (approved):

PR #128, opened 2024-08-26
refactor: merge user config

Timeless0911:
## Summary

do not need to `mergeRsbuildConfig` twice

## Checklist

### DIFF
--- a/packages/core/tests/__snapshots__/config.test.ts.snap
+++ b/packages/core/tests/__snapshots__/config.test.ts.snap
@@ -7,6 +7,7 @@ exports[`Should compose create Rsbuild config correctly > Merge Rsbuild config 1
       "dev": {
         "progressBar": false,
       },
+      "format": "esm",
       "mode": "production",
       "output": {
         "distPath": {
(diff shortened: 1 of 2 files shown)


packages/core/tests/__snapshots__/config.test.ts.snap
```diff
@@ -7,6 +7,7 @@ exports[`Should compose create Rsbuild config correctly > Merge Rsbuild config 1
       "dev": {
         "progressBar": false,
       },
+      "format": "esm",
```
fi3ework, line 10: We're adding unknown field to Rsbuild config here, could we prevent this?

fi3ework (approved):

packages/core/tests/__snapshots__/config.test.ts.snap
Timeless0911: Just add format this time, and other unknown field is already there for severral days. Rsbuild inspect seems to filter them in display.

fi3ework: > and other unknown field is already there for severral days

Did we? The seems not in the snapshot at the moment.

Timeless0911: because we only add format in our unit test

PR #135, opened 2024-08-27
feat: modify merge config order

Timeless0911:
## Summary

The priorities from high to low are as follows:

1. `userConfig`: users can configure any Rsbuild and Rspack config
2. `libRsbuildConfig`: the configuration that we compose from Rslib unique config and userConfig from 1
3. `internalRsbuildConfig`: the built-in best practice Rsbuild configuration we provide in Rslib

We should state in the document that the built-in configuration should not be changed optionally

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -660,7 +660,15 @@ export async function composeCreateRsbuildConfig(
 
     return {
       format: libConfig.format!,
+      // The merge order represents the priority of the configuration
+      // The priorities from high to low are as follows:
+      // 1 - userConfig: users can configure any Rsbuild and Rspack config
+      // 2 - libRsbuildConfig: the configuration that we compose from Rslib unique config and userConfig from 1
+      // 3 - internalRsbuildConfig: the built-in best practice Rsbuild configuration we provide in Rslib
+      // We should state in the document that the built-in configuration should not be changed optionally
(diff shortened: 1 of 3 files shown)


packages/core/src/config.ts
```diff
@@ -660,7 +660,17 @@ export async function composeCreateRsbuildConfig(
...
+      // The merge order represents the priority of the configuration
+      // The priorities from high to low are as follows:
+      // 1 - userConfig: users can configure any Rsbuild and Rspack config
+      // 2 - libRsbuildConfig: the configuration that we compose from Rslib unique config and userConfig from 1
+      // 3 - internalRsbuildConfig: the built-in best practice Rsbuild configuration we provide in Rslib
+      // We should state in the document that the built-in configuration should not be changed optionally
+      // In compose process of 2, we may read some config from 1, and reassemble the related config,
+      // so before final mergeRsbuildConfig, we reset some specified fields
       config: mergeRsbuildConfig(
+        internalRsbuildConfig,
```
fi3ework, line 672: `internalRsbuildConfig` is just some fixed configs that even do not get involved into compose. I think we could rename it to `staticConifg`, `constantConfig` or `fixedConfig`. We can change it in future PRs alongside.

fi3ework (approved):

Issue #141, opened 2024-08-29
Priority between `output.externals` and `resolve.alias` config


Timeless0911:
### Details

```ts
alias: {
  react: 'react-native',
},
```

If we want to make `react` alias to `react-native`, and the **alias** behaviour will affected by **`externals`** config

## bundle

In bundle mode, the `autoExternal` is default set to true, so react will be externals to `^react($|\\/|\\\\)`.

So the alias config is invalid after externals set.

We should set `autoExternal: false` or set `output.externals` to make alias work

```ts
output: {
  externals: {
    react: 'react-native',
  },
},
```

## bundleless

In bundleless mode, we use externals like below to generate bundleless output.

https://github.com/web-infra-dev/rslib/blob/4b5f674789059bea3456e1105cb3cf9b051912d7/packages/core/src/config.ts#L506

So the react request is callback here, so alias config is invalid after externals set.

We need to set `output.externals` to make alias work, and the alias config turn out to be logic that won't be executed

```ts
output: {
  externals: {
    react: 'react-native',
  },
},
```

Timeless0911:
Add docs in `source.alias` and 	`output.external` to explain the relationship between each other.

- bundle: use `source.alias` to alias, note the effect of `autoExternal`
- bundleless: use `output.external` to alias

Timeless0911:
https://lib.rsbuild.dev/config/rsbuild/source#sourcealias

PR #143, opened 2024-08-29
test: add resolve tests

Timeless0911:
## Summary

add entry tests to align with tests of Modern.js Module `resolve` config

alias tests is not added since https://github.com/web-infra-dev/rslib/issues/141, we may add externals test in the future

## Checklist


e2e/cases/resolve/__snapshots__/index.test.ts.snap
```diff
@@ -0,0 +1,72 @@
...
+    __webpack_require__.r = function(exports) {
+        if ('undefined' != typeof Symbol && Symbol.toStringTag) Object.defineProperty(exports, Symbol.toStringTag, {
+            value: 'Module'
+        });
+        Object.defineProperty(exports, '__esModule', {
+            value: true
+        });
+    };
+})();
+/************************************************************************/ var __webpack_exports__ = {};
```
fi3ework, line 66: emmm, so much runtime code. It's set to `esm` already, but it seems not using the optimized concatenated modules runtime, which should be a bug.

e2e/cases/resolve/false/index.ts
```diff
@@ -0,0 +1,3 @@
+import jjj from './browser-false/util';
+
+console.log('jjj:', jjj); // the value of `jjj` should `{}`
```
fi3ework, line 3: `jjj` looks weird, we could change to `value` or `foo` to fix the bad pattern from modern module.

e2e/cases/resolve/with-condition-exports/.gitignore
```diff
@@ -0,0 +1,2 @@
+!node_modules
```
fi3ework, line 1: useless?

e2e/cases/resolve/with-main-fields/.gitignore
```diff
@@ -0,0 +1,2 @@
+!node_modules
```
fi3ework, line 1: useless?

e2e/cases/resolve/with-main-fields/node_modules/lib2/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "browser": "src/browser.ts",
+  "main": "src/main.ts",
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
```
fi3ework, line 12: I would prefer to use `MIT` as it's more general.

e2e/scripts/shared.ts
```diff
@@ -34,7 +34,7 @@ export function generateBundleCjsConfig(config: LibConfig = {}): LibConfig {
   return mergeConfig(cjsBasicConfig, config)!;
 }
 
-type FormatType = 'esm' | 'cjs';
+type FormatType = 'esm' | 'cjs' | string;
```
fi3ework, line 37: `FormatType` should be enumed.

e2e/cases/resolve/with-main-fields/.gitignore
Timeless0911: This file is aim to keep node_modules folder exist to test resolve logic

e2e/scripts/shared.ts
fi3ework: Could change to

```ts
type Format = 'esm' | 'cjs'
type EnvName = Format | `${Format}${number}`
```

e2e/cases/resolve/__snapshots__/index.test.ts.snap
Timeless0911:
```
var __create = Object.create;
var __defProp = Object.defineProperty;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __getProtoOf = Object.getPrototypeOf;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __commonJS = (cb, mod) => function __require() {
  return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
};
var __copyProps = (to, from, except, desc) => {
  if (from && typeof from === "object" || typeof from === "function") {
    for (let key of __getOwnPropNames(from))
      if (!__hasOwnProp.call(to, key) && key !== except)
        __defProp(to, key, { get: () => from[key], enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable });
  }
  return to;
};
var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__getProtoOf(mod)) : {}, __copyProps(
  // If the importer is in node compatibility mode or this is not an ESM
  // file that has been converted to a CommonJS file using a Babel-
  // compatible transform (i.e. "__esModule" has not been set), then set
  // "default" to the CommonJS "module.exports" for node compatibility.
  isNodeMode || !mod || !mod.__esModule ? __defProp(target, "default", { value: mod, enumerable: true }) : target,
  mod
));

// resolve-false:/empty-stub
var require_empty_stub = __commonJS({
  "resolve-false:/empty-stub"(exports, module2) {
    module2.exports = {};
  }
});

// index.ts
var import_util = __toESM(require_empty_stub());
console.log("jjj:", import_util.default);
```

esbuild output is this

Timeless0911: resolve false should generate a not-exist module, which may import such runtime code to generate `__webpack_require__`

e2e/cases/resolve/__snapshots__/index.test.ts.snap
```diff
@@ -51,22 +51,13 @@ function __webpack_require__(moduleId) {
...
-    };
-})();
-/************************************************************************/ var __webpack_exports__ = {};
-__webpack_require__.r(__webpack_exports__);
-/* harmony import */ var _browser_false_util__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(\\"?69d8\\");
-/* harmony import */ var _browser_false_util__WEBPACK_IMPORTED_MODULE_0___default = /*#__PURE__*/ __webpack_require__.n(_browser_false_util__WEBPACK_IMPORTED_MODULE_0__);
-console.log('foo:', _browser_false_util__WEBPACK_IMPORTED_MODULE_0___default()); // the value of \`foo\` should \`{}\`
+/************************************************************************/ // CONCATENATED MODULE: ./bar.ts
+const bar = 'bar';
+// EXTERNAL MODULE: <ROOT>/e2e/cases/resolve/false/./browser-false/util (ignored)
```
Timeless0911, line 56: @Timeless0911 added an extra module

fi3ework: Tracks the edge case in https://github.com/web-infra-dev/rslib/issues/149.

fi3ework (approved):

PR #144, opened 2024-08-29
fix: warn should handles all externals config

### LINKED ISSUE #142: [Bug]: should not warn when commonjs request is not external
### Details

```ts
// entry.ts
console.log(require('lib1').value);
```

```js
// lib1
module.exports.value = 'lib1 cjs'
```

When bundle esm, it will cause warning in log
image

fi3ework:
## Summary

Fix #142.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -88,6 +88,108 @@ export async function loadConfig({
   return content as RslibConfig;
 }
 
+const composeExternalsWarnConfig = (
+  format: Format,
+  ...externalsArray: NonNullable<RsbuildConfig['output']>['externals'][]
+): RsbuildConfig => {
+  if (format !== 'esm') {
+    return {};
+  }
+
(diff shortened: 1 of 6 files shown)


packages/core/src/config.ts
```diff
@@ -88,6 +88,108 @@ export async function loadConfig({
...
+        ({ request, dependencyType, contextInfo }: any, callback: any) => {
+          let externalized = false;
+          const _callback = (matched?: true) => {
+            if (matched) {
+              externalized = true;
+            }
+          };
+
+          if (contextInfo.issuer && dependencyType === 'commonjs') {
+            matchUserExternals(externals, request, _callback);
```
Timeless0911, line 181: Why not return boolean directly insteadof using callbacl function?

fi3ework: It's somehow a complex logic, just to port webpack's logic and edit as less as possible.

Timeless0911 (approved):

PR #145, opened 2024-08-30
feat: setup `create-rslib` package

chenjiahan:
## Summary

Setup the `create-rslib` package, it is implemented based on [create-rstack](https://github.com/rspack-contrib/create-rstack).

Screenshot 2024-08-30 at 11 15 48

Note that this PR only adds two example templates for JS / TS, and we should provide different templates for Node.js library / React Components / Vue Components in the future.

## Related Links

- https://github.com/rspack-contrib/create-rstack
- https://github.com/web-infra-dev/rslib/issues/46

## Checklist


Timeless0911 (approved):
LGTM

PR #163, opened 2024-08-31
fix: syntax could handle combined pattern

fi3ework:
## Summary

In the past, for syntax, `[es5]` and `es5` were different, the former would be treated as browserslist processing while the latter is a valid esX version. This PR has removed this restriction, now `[es5]` and `es5` are equivalent, and esX can participate in the target syntax grammar, such as `['es2015', 'Chrome 123']`. Array-form syntax now is the superset of browserslist.

## Checklist

### DIFF
--- a/e2e/cases/syntax/__snapshots__/index.test.ts.snap
+++ b/e2e/cases/syntax/__snapshots__/index.test.ts.snap
@@ -34,3 +34,14 @@ function bar() {} /*#__PURE__*/
 export { Foo };
 "
 `;
+
+exports[`should downgrade class private method with output.syntax config 2`] = `
+"class Foo {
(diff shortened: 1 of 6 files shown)


fi3ework:
Is it weird to extending browserslist query with `esX`?

Timeless0911:
> Is it weird to extending browserslist query with `esX`?

It feels OK as long as the final result we calculate meets the user's semantics.

Timeless0911:
Should we use `syntax` directly instead of `output.syntax`?

fi3ework:
> Should we use syntax directly instead of output.syntax?

In a separated PR.

Timeless0911 (approved):

PR #164, opened 2024-09-01
fix: could accept minify from user

fi3ework:
## Summary

Use the minify option when user provided and stop merging the default options. If not provided, use the default options.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -247,6 +247,41 @@ export const composeAutoExternalConfig = (options: {
     : {};
 };
 
+export function composeMinifyConfig(
+  minify: NonNullable<RsbuildConfig['output']>['minify'],
+): RsbuildConfig {
+  if (minify !== undefined) {
+    // User's minify configuration will be merged afterwards.
+    return {};
+  }
+
+  // When minify is not specified, Rslib will use a sane default for minify options.
+  // The default options will only perform dead code elimination and unused code elimination.
+  return {
+    output: {
+      minify: {
+        js: true,
+        css: false,
+        jsOptions: {
+          minimizerOptions: {
+            mangle: false,
+            minify: false,
+            compress: {
+              defaults: false,
+              unused: true,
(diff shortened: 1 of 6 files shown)


packages/core/src/config.ts
```diff
@@ -247,6 +247,44 @@ export const composeAutoExternalConfig = (options: {
     : {};
 };
 
+export function composeMinifyConfig(
+  minify: NonNullable<RsbuildConfig['output']>['minify'],
+): RsbuildConfig {
+  if (minify !== undefined) {
+    return {
```
Timeless0911, line 254: Should we delete use's `output.minify` config before final config merge process. Or we can return `{}` here, and final merge process will ship user's minify config

fi3ework: Nice catch, I returned `{}` here.

packages/core/src/config.ts
```diff
@@ -247,6 +247,41 @@ export const composeAutoExternalConfig = (options: {
...
+    // User's minify configuration will be merged afterwards.
+    return {};
+  }
+
+  // When minify is not specified, Rslib will use a sane default for minify options.
+  // The default options will only perform dead code elimination and unused code elimination.
+  return {
+    output: {
+      minify: {
+        js: true,
```
chenjiahan, line 263: Why do need minify JS files by default

chenjiahan: Libraries usually ship uon-minified code to npm.

fi3ework: By default, it's not minified. Adding `minify: true` is required as we want to perform some tree-shaking related compress options by default (https://github.com/web-infra-dev/rslib/blob/d68915dd2294a0a0d2e767e9ed672c25fdb2f6a8/packages/core/src/config.ts#L288-L294).

chenjiahan: get

Timeless0911 (approved):

PR #169, opened 2024-09-02
feat!: move config `output.syntax` to `syntax`

fi3ework:
## Summary

We try to flatten the fields under `lib` as much as possible instead of nesting them.

## Checklist


chenjiahan (approved):
LGTM 👍

PR #173, opened 2024-09-04
chore: not use recursive omit

Timeless0911:
## Summary

`omitDeep` will break some element like 
```TS
tools: {
  rspack: {
    plugins: [
      new rspack.BannerPlugin({
        banner: 'hello world',
        footer: true,
        stage: 10000,
      }),
    ],
  },
},
```

## Checklist


fi3ework (approved):
Hope we won't need this TS snippet again one day. :P

PR #189, opened 2024-09-09
feat: set legacy decorators version when `experimentalDecorators` enabled

Timeless0911:
## Summary

- use tsconfck to find and parse tsconfig file
- set legacy decorators version when `experimentalDecorators` enabled
- add `tsconfigPath` test cases
- prebundle `rslog`

## Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -51,6 +51,7 @@
     "prebundle": "1.2.2",
     "rslib": "npm:@rslib/core@0.0.5",
     "rslog": "^1.2.3",
+    "tsconfck": "3.1.3",
     "typescript": "^5.5.4"
   },
   "peerDependencies": {
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -10,7 +10,6 @@ import {
   rspack,
 } from '@rsbuild/core';
 import glob from 'fast-glob';
-import ts from 'typescript';
 import {
   DEFAULT_CONFIG_NAME,
   DEFAULT_EXTENSIONS,
@@ -41,7 +40,7 @@ import {
 } from './utils/helper';
 import { logger } from './utils/logger';
 import { transformSyntaxToBrowserslist } from './utils/syntax';
-import { getTsconfigCompilerOptions } from './utils/tsconfig';
+import { loadTsconfig } from './utils/tsconfig';
 
 /**
  * This function helps you to autocomplete configuration types.
@@ -363,12 +362,12 @@ export function composeBannerFooterConfig(
 }
 
 export function composeDecoratorsConfig(
-  options: ts.CompilerOptions,
+  compilerOptions?: Record<string, any>,
   version?: NonNullable<
     NonNullable<RsbuildConfig['source']>['decorators']
   >['version'],
 ): RsbuildConfig {
-  if (version || !options.experimentalDecorators) {
+  if (version || !compilerOptions?.experimentalDecorators) {
     return {};
   }
 
@@ -811,15 +810,10 @@ const composeExternalHelpersConfig = (
 async function composeLibRsbuildConfig(config: LibConfig, configPath: string) {
   const rootPath = dirname(configPath);
   const pkgJson = readPackageJson(rootPath);
-  let compilerOptions: ts.CompilerOptions = {};
-  const tsconfigPath = ts.findConfigFile(
+  const { compilerOptions } = await loadTsconfig(
     rootPath,
-    ts.sys.fileExists,
     config.source?.tsconfigPath,
   );
-  if (tsconfigPath) {
-    compilerOptions = getTsconfigCompilerOptions(tsconfigPath);
-  }
 
   const {
     format,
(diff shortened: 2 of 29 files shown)


fi3ework (commented):
I just recalled https://github.com/web-infra-dev/modern.js/issues/5834.

How does esbuild do? See https://github.com/evanw/esbuild/blob/main/CHANGELOG.md#0210:

> Note that this feature is different than the existing "TypeScript experimental decorators" feature that esbuild already implements. It uses similar syntax but behaves very differently, and the two are not compatible (although it's sometimes possible to write decorators that work with both). TypeScript experimental decorators will still be supported by esbuild going forward as they have been around for a long time, are very widely used, and let you do certain things that are not possible with JavaScript decorators (such as decorating function parameters). By default esbuild will parse and transform JavaScript decorators, but you can tell esbuild to parse and transform TypeScript experimental decorators instead by setting "experimentalDecorators": true in your tsconfig.json file.

For projects that have `experimentalDecorator` enabled in tsconfig, should we automatically switch to `legacyDecorator` (and `emitDecoratorMetadata`)? Additionally, when we perform checks between configurations later on, `legacyDecorator` should also check whether it is compatible with `decorators.version`.

In create-rslib, I think we're good to use the new version of decorator by default.

Timeless0911:
https://github.com/web-infra-dev/rsbuild/blob/01a8b34143b4957225186e9b80b050103582e546/packages/core/src/plugins/swc.ts#L250-L273

The Rsbuild logic is above, and we should just do a tsconfig detect, and if users config `experimentalDecorator: true`, we automatically inject legacy decorator Rsbuild config.

fi3ework:
> [web-infra-dev/rsbuild@`01a8b34`/packages/core/src/plugins/swc.ts#L250-L273](https://github.com/web-infra-dev/rsbuild/blob/01a8b34143b4957225186e9b80b050103582e546/packages/core/src/plugins/swc.ts#L250-L273)
> 
> The Rsbuild logic is above, and we should just do a tsconfig detect, and if users config `experimentalDecorator: true`, we automatically inject legacy decorator Rsbuild config.

Should this be handled by Rsbuild? I think it also applies to app building.

Timeless0911:
> > [web-infra-dev/rsbuild@`01a8b34`/packages/core/src/plugins/swc.ts#L250-L273](https://github.com/web-infra-dev/rsbuild/blob/01a8b34143b4957225186e9b80b050103582e546/packages/core/src/plugins/swc.ts#L250-L273)
> > The Rsbuild logic is above, and we should just do a tsconfig detect, and if users config `experimentalDecorator: true`, we automatically inject legacy decorator Rsbuild config.
> 
> Should this be handled by Rsbuild? I think it also applies to app building.

Rsbuild does not want to be associated with tsconfig.json fields.

packages/core/src/config.ts
```diff
@@ -10,6 +10,7 @@ import {
   rspack,
 } from '@rsbuild/core';
 import glob from 'fast-glob';
+import ts from 'typescript';
```
chenjiahan, line 13: this will break JS projects

packages/core/src/config.ts
```diff
@@ -360,6 +362,25 @@ export function composeBannerFooterConfig(
...
+  >['version'],
+): RsbuildConfig {
+  if (version || !options.experimentalDecorators) {
+    return {};
+  }
+
+  return {
+    source: {
+      decorators: {
+        version: 'legacy',
```
chenjiahan, line 378: why defaults to legacy

packages/core/src/config.ts
fi3ework: We could use tsconfck to look up tsconfig, which is also used by Vite.

packages/core/src/config.ts
Timeless0911: This is the return value of `1&&2`
1. no `source.decorators.version` config
2. `experimentalDecorators`: true in tsconfig

Timeless0911: Maybe I should write on the contrary

```ts
  if (!version && options.experimentalDecorators) {
    return {
    source: {
      decorators: {
        version: 'legacy',
      },
    },
  };
  }

  return {};
```

chenjiahan: My mistake, the current implementation is ok

packages/core/src/config.ts
Timeless0911: I use `tsconfck` instead,  more details can be implemented when dealing no config feature.

packages/core/package.json
```diff
@@ -51,6 +51,7 @@
     "prebundle": "1.2.2",
     "rslib": "npm:@rslib/core@0.0.5",
     "rslog": "^1.2.3",
+    "tsconfck": "3.1.3",
```
fi3ework, line 54: Do we need to precompile it or using the `autoExternal` feature instead? I don't know how to choose between these two now TBH 😂.

fi3ework: Got it, tsconfck is an aggressive ESM package without main. I think prebundle should also support `import.meta.resolve()` to resolve ESM entry as well in the future. I also found that rslog is not prebundled, it that on purpose? 🤔

Timeless0911: Forgot to add rslog before.

fi3ework (approved):

PR #192, opened 2024-09-11
ci: only trigger docs preview on demand

fi3ework:
## Summary

Don't create preview documents for every pull request, it's noisy and wastes resources. Only trigger them when the documentation is relevant.

https://docs.netlify.com/configure-builds/ignore-builds/#mimic-default-behavior:~:text=The%20following%20ignore%20command%20example%20adapts%20the%20default%20behavior%20so%20that%20the%20build%20proceeds%20only%20if%20there%20are%20changes%20within%20the%20blog%2D1%20or%20common%20directories.

## Checklist


Timeless0911 (approved):
👍

Issue #209, opened 2024-09-18
[Bug]: should not set `target: 'es5'` when using `syntax`


chenjiahan:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 67.81 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 128.0.6613.138
    Safari: 17.6
  npmPackages:
    @rslib/core: 0.0.6 => 0.0.6
```

### Details

Rslib should not set `target: 'es5'` when using `syntax`, this will increase the bundle size.

### Reproduce link

https://github.com/web-infra-dev/rslib/blob/ed9fe5b0760822e703d9703d5ec31dc86ead2e5f/packages/core/src/config.ts#L569

### Reproduce Steps

See https://github.com/web-infra-dev/rsbuild/pull/3502

PR #213, opened 2024-09-18
fix: derive `rspack.target` from `libConfig.syntax`

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/209.

- only webpack runtime related code will be affected.
- need to manually assign `{type}loading` as it derive the value from `target` by default, which will lead to unimpelmented `universal` type.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -560,18 +560,17 @@ const composeSyntaxConfig = (
   // Defaults to ESNext, Rslib will assume all of the latest JavaScript and CSS features are supported.
 
   if (syntax) {
+    const resolvedBrowserslist = transformSyntaxToBrowserslist(syntax, target);
     return {
       tools: {
         rspack: (config) => {
-          // TODO: Rspack should could resolve `browserslist:{query}` like webpack.
-          // https://webpack.js.org/configuration/target/#browserslist
-          // Using 'es5' as a temporary solution for compatibility.
-          config.target = ['es5'];
-          return config;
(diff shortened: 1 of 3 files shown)


fi3ework:
https://github.com/web-infra-dev/rslib/blob/main/netlify.toml is not working 🤨.

Timeless0911:
> [`main`/netlify.toml](https://github.com/web-infra-dev/rslib/blob/main/netlify.toml?rgh-link-date=2024-09-18T07%3A34%3A48Z) is not working 🤨.

I rerun it, it seems work. Try to force push again to verify.

fi3ework:
https://app.netlify.com/sites/rslib/deploys

Google Chrome 2024-09-18 15 55 56

Weird, the below one was triggered by this PR, which is canceled by as expected. The above one still finish the preview build and deploy. We could test in following PRs, PLAT this PR.

packages/core/src/config.ts
```diff
@@ -725,6 +724,11 @@ const composeTargetConfig = (
         tools: {
           rspack: {
             target: ['web'],
+            output: {
```
Timeless0911, line 727: What are the problems with inferring based on target and chunkformat?

fi3ework: > need to manually assign {type}loading as it derive the value from target by default, which will lead to unimpelmented universal type.

Timeless0911 (approved):

PR #216, opened 2024-09-18
feat: support `import.meta.url` shim for CJS

fi3ework:
## Summary

This Rsbuild plugin will shim `import.meta.url` for CommonJS modules.
- Replace `import.meta.url` with `importMetaUrl`.
- Inject `importMetaUrl` to the end of the module (can't inject at the beginning because of `"use strict";`).
This is a short-term solution, and we hope to provide built-in polyfills like `node.__filename` on Rspack side.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -15,6 +15,7 @@ import {
   DEFAULT_EXTENSIONS,
   SWC_HELPERS,
 } from './constant';
+import { pluginCjsShim } from './plugins/cjsShim';
 import type {
   AutoExternal,
   BannerAndFooter,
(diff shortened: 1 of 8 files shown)


packages/core/src/config.ts
```diff
@@ -456,8 +457,16 @@ const composeFormatConfig = (format: Format): RsbuildConfig => {
       };
     case 'cjs':
       return {
+        plugins: [pluginCjsShim()],
```
Timeless0911, line 460: Why using a plugin for this feature?

fi3ework: Easy to remove in the future, and implement is a plugin form is way more decouple from other implementations and more easy to scale.

Timeless0911 (approved):

Issue #233, opened 2024-09-25
[Feature]: Support for `createRequire()`


chenjiahan:
### What problem does this feature solve?

- Input:

```js
import { createRequire } from 'node:module';

const require = createRequire(import.meta.url);

require.resolve('stylus-loader');
```

- Current output of Rslib 0.0.7:

```js
// mjs
import * as __WEBPACK_EXTERNAL_MODULE_node_module__ from "node:module";
const src_require = (0, __WEBPACK_EXTERNAL_MODULE_node_module__.createRequire)(__rslib_import_meta_url__);
src_require.resolve('stylus-loader');

// cjs
var __webpack_exports__ = {};
const external_node_module_namespaceObject = require("node:module");
const src_require = (0, external_node_module_namespaceObject.createRequire)(__rslib_import_meta_url__);
src_require.resolve('stylus-loader');
var __webpack_export_target__ = exports;
for(var i in __webpack_exports__)__webpack_export_target__[i] = __webpack_exports__[i];
if (__webpack_exports__.__esModule) Object.defineProperty(__webpack_export_target__, '__esModule', {
    value: true
});
var __rslib_import_meta_url__ = /*#__PURE__*/ function() {
    return 'undefined' == typeof document ? new (require('url'.replace('', ''))).URL('file:' + __filename).href : document.currentScript && document.currentScript.src || new URL('main.js', document.baseURI).href;
}();
```

### What does the proposed API look like?

The expected output should be similar to `tsup`:

```js
// mjs
import { createRequire } from "node:module";
var require2 = createRequire(import.meta.url);
require2.resolve("stylus-loader");

// cjs
var getImportMetaUrl = () => typeof document === "undefined" ? new URL(`file:${__filename}`).href : document.currentScript && document.currentScript.src || new URL("main.js", document.baseURI).href;
var importMetaUrl = /* @__PURE__ */ getImportMetaUrl();
var import_node_module = require("module");
var require2 = (0, import_node_module.createRequire)(importMetaUrl);
require2.resolve("stylus-loader");
```

PR #234, opened 2024-09-25
fix: CJS shims should not affect other environments

fi3ework:
## Summary

IIUC, it could fix https://github.com/web-infra-dev/rslib/issues/233. cc @chenjiahan , is this the bug that this PR fixes, as described in that issue?

## Checklist

### DIFF
--- a/e2e/cases/shims/index.test.ts
+++ b/e2e/cases/shims/index.test.ts
@@ -1,4 +1,5 @@
 import { join } from 'node:path';
+import { describe } from 'node:test';
 import { buildAndGetResults } from '@e2e/helper';
 import { expect, test } from 'vitest';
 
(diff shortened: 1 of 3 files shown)


e2e/cases/shims/index.test.ts
```diff
@@ -1,4 +1,5 @@
 import { join } from 'node:path';
+import { describe } from 'node:test';
```
Timeless0911, line 2: why from node test

fi3ework: 😅 intellisense is not intelligent enough

Timeless0911 (approved):

chenjiahan:
LGTM, I will test the new version in Rsbuild.

Issue #204, opened 2024-09-14
[Bug]: external should handle `require.resolve`


SoonIter:
### Version

```sh
rslib 0.0.5
```

### Details

<img width="935" alt="image" src="https://github.com/user-attachments/assets/01297f0e-0fea-47a1-b5c6-e6dfeb6d45dc">

### Reproduce link

none

### Reproduce Steps

none

PR #249, opened 2024-09-30
fix: preserve require.resolve / dynamic require / dynamic import

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/204.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -432,6 +432,14 @@ export async function createConstantRsbuildConfig(): Promise<RsbuildConfig> {
 }
 
 const composeFormatConfig = (format: Format): RsbuildConfig => {
+  const jsParserOptions = {
+    importMeta: false,
+    requireResolve: false,
+    requireDynamic: false,
+    requireAsExpression: false,
+    importDynamic: false,
+  };
+
   switch (format) {
     case 'esm':
       return {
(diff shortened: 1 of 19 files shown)


packages/core/src/config.ts
```diff
@@ -446,13 +454,12 @@ const composeFormatConfig = (format: Format): RsbuildConfig => {
...
               parser: {
-                javascript: {
-                  importMeta: false,
-                },
+                javascript: jsParserOptions,
               },
             },
             optimization: {
               concatenateModules: true,
+              sideEffects: 'flag',
```
Timeless0911, line 462: What is this for?

fi3ework: https://webpack.js.org/configuration/optimization/#optimizationsideeffects, if a dep didn't marking `sideEffects: false` in package.json, then it will not be tree shaken.

**Why we need to change it to `'flag'`:**

to make the modules in the source code didn't get bail out from https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/normal_module.rs#L525-L530 by when process concatenation. See the deleted code in snapshot.

**What does it cost?**

When bundling other modules, tree shaking performance may not be optimal. For cases sensitive to tree shaking, such as ESM, we typically avoid bundling the dependencies.

In the long run, it would be better to add a configuration that allows for fine-grained control over the level of tree shaking applied to specific modules.

Timeless0911 (approved):

Issue #254, opened 2024-10-07
[Bug]: `require` call appears in ESM output when there’s a dynamic import.


xc2:
### Version

```sh
  System:
    OS: macOS 15.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 72.34 MB / 32.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Browsers:
    Chrome: 129.0.6668.90
    Safari: 18.1
  npmPackages:
    @rslib/core: 0.0.8 => 0.0.8
```

### Details

I'm using dynamic import in a Node.js package and bundling it into an esm output. The code looks like:

```ts
export async function reproduction(input: string) {
  const { default: fnv1a } = await import(
    /* webpackChunkName: "_sindresorhus_fnv1a" */ "@sindresorhus/fnv1a"
  );
  console.log(`

      fnv1a hash of "${input}" is "${fnv1a(input, { size: 32 }).toString(16)}"

`);
}
```

Note: `@sindresorhus/fnv1a` here is introduced as `devDependencies` as I'd like it bundled into my package.

`rslib.config.ts` is deadly simple

```ts
import { defineConfig } from "@rslib/core";

export default defineConfig({
  source: {
    entry: {
      index: "./src/index.ts",
    },
  },
  output: {
    target: "node",
    sourceMap: {
      js: "source-map",
    },
    minify: { js: false },
    distPath: {
      root: "./dist",
    },
  },
  lib: [
    {
      format: "esm",
      dts: false,
      autoExtension: false,
      output: {
        filename: {
          js: "[name].mjs",
        },
      },
    },
  ],
});
```

The usage is like

```ts
import { reproduction } from "cacheable-crawlee";

await reproduction("it works");
```

which raises

```sh
file:///Users/kfll/Developer/crawlee-cache/dist/index.mjs:97
                        installChunk(require("./" + __webpack_require__.u(chunkId)));
                        ^

ReferenceError: require is not defined in ES module scope, you can use import instead
    at __webpack_require__.f.require (file:///Users/kfll/Developer/crawlee-cache/dist/index.mjs:97:4)
    at file:///Users/kfll/Developer/crawlee-cache/dist/index.mjs:48:30
    at Array.reduce (<anonymous>)
    at __webpack_require__.e (file:///Users/kfll/Developer/crawlee-cache/dist/index.mjs:47:38)
[...]

PR #256, opened 2024-10-08
fix: use ESM chunkLoading for "esm" format

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/254.

## Checklist


fi3ework:
@Timeless0911 Same, but I don't think this will work for typical web scenes. Due to https://github.com/webpack/webpack/issues/15697, the dynamic chunk can't be bundled it can't be statically resolved.

Timeless0911 (approved):

PR #273, opened 2024-10-10
chore: use 0.0.10 to build self

fi3ework:
## Checklist

### DIFF
--- a/tests/package.json
+++ b/tests/package.json
@@ -13,7 +13,7 @@
   "devDependencies": {
     "@codspeed/vitest-plugin": "^3.1.1",
     "@playwright/test": "1.47.2",
-    "@rsbuild/core": "1.0.7",
+    "@rsbuild/core": "1.0.12",
     "@rsbuild/plugin-less": "^1.0.1",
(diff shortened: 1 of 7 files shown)


fi3ework:
We need to speed up CI, especially benchmark. Considering disable dts when building, we need to pass inlined config to merge with the file config first.

Timeless0911:
The lockfile change seems a little extra, pnpm update may cause resolution again.

fi3ework:
Reverted part of the lockfile.

tests/package.json
```diff
@@ -13,7 +13,7 @@
   "devDependencies": {
     "@codspeed/vitest-plugin": "^3.1.1",
     "@playwright/test": "1.47.2",
-    "@rsbuild/core": "1.0.7",
+    "@rsbuild/core": "1.0.12",
```
Timeless0911, line 16: Why this not be bumped by renovate

Timeless0911: image
we should modify this and the `e2e/cases` is deprecated now.

Timeless0911 (approved):

Issue #283, opened 2024-10-12
[Bug]: Rslib failed to report the type errors


chenjiahan:
### Version

Rslib 0.0.11

### Details

See:

![Image](https://github.com/user-attachments/assets/f588e2cd-be05-4965-bb2b-10e16f597e4b)

### Reproduce link

https://github.com/web-infra-dev/rsbuild/pull/3691

### Reproduce Steps

1. cd packages/plugin-babel
2. pnpm build

Timeless0911:
This may be the difference of Typescript Compiler API, running `pnpm build --watch` will throw these errors.

- `pnpm build`: `ts.createProgram`
- `pnpm build --watch`: `ts.createSemanticDiagnosticsBuilderProgram`

PR #288, opened 2024-10-14
fix: pass in projectReferences to avoid missing some error reports

Timeless0911:
## Summary

fix: #283 

## Checklist

### DIFF
--- a/.changeset/proud-sheep-end.md
+++ b/.changeset/proud-sheep-end.md
@@ -1,0 +1,5 @@
+---
+'@rslib/core': patch
+---
+
+release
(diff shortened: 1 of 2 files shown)


fi3ework:
Do we need to add rest fields alongside, are they useful?

image

Timeless0911:
> Do we need to add rest fields alongside, are they useful?
> 
> image

According to Typescript source code, it seems they are not useful yet.

```ts
interface CreateProgramOptions {
    rootNames: readonly string[];
    options: CompilerOptions;
    projectReferences?: readonly ProjectReference[];
    host?: CompilerHost;
    oldProgram?: Program;
    configFileParsingDiagnostics?: readonly Diagnostic[];
}
```

fi3ework (approved):

PR #293, opened 2024-10-15
perf(core): optimize the resolveConfigPath

SoonIter:
## Summary

by the way, test the codSpeed

## Checklist


fi3ework:
why closed, is there no perf gain?

PR #299, opened 2024-10-16
feat: support "umd" format

fi3ework:
## Summary

TODO:

`umdModuleName`  lacks some advanced features:

1. non string type
2. auto transform to camel case

We might implement them after understand why they're supported in https://modernjs.dev/module-tools/api/config/build-config.html#umdglobals.

---

update: 

1. Don't need to support `auto transform to camel case`, it was a limitation when implement with swc.
2. `non string type` might support it.

## Checklist


packages/core/src/types/config/index.ts
```diff
@@ -80,6 +80,7 @@ export interface LibConfig extends RsbuildConfig {
   footer?: BannerAndFooter;
   shims?: Shims;
   dts?: Dts;
+  umdName?: string;
```
Timeless0911, line 83: add to https://github.com/web-infra-dev/rslib/blob/main/packages/core/src/config.ts#L1100

tests/e2e/react-component/package.json
```diff
@@ -2,8 +2,14 @@
   "name": "react-component-e2e",
   "version": "1.0.0",
   "private": true,
+  "scripts": {
+    "dev:bundle": "../../node_modules/.bin/rsbuild dev --environment=bundle",
+    "dev:bundle-false": "../../node_modules/.bin/rsbuild dev --environment=bundleFalse",
+    "dev:umd": "../../node_modules/.bin/rsbuild dev --environment=umd"
```
Timeless0911, line 8: why not add rsbuild to devdeps in this package.json

fi3ework: It's using @rsbuild/core in `../../` to start by the test runner.

packages/core/src/types/config/index.ts
fi3ework: I changed the `omit` function, so it's required to add new keys here in the future.

tests/e2e/react-component/package.json
Timeless0911: using relative path to call bin file is a bit ugly

fi3ework: Could refactor in other PR.

Timeless0911 (approved):

Issue #284, opened 2024-10-12
[Bug]: Should not lower function default parameters when using `syntax: 'es2021'


chenjiahan:
### Version

Rslib 0.0.11

### Details

- Rslib config:

```js
export default {
  lib: [
    {
      format: 'esm',
      syntax: 'es2021',
    },
  ],
  output: {
    target: 'node',
  },
};
```

- Input:

```ts
export function foo(options: unknown = {}): void {
  console.log(options);
}
```

- Current output:

```js
function foo() {
    let options = arguments.length > 0 && void 0 !== arguments[0] ? arguments[0] : {};
    console.log(options);
}
export { foo };
```

- Expected output:

```js
function foo(options = {}) {
    console.log(options);
}
export { foo };
```

### Reproduce link

I found this issue in some Rsbuild packages

### Reproduce Steps

/

PR #301, opened 2024-10-16
fix: correct ECMA version browserslist mapping

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/284.

## Checklist


chenjiahan (commented):
Most build tools support `es6` as an alias to `es2015`, can we consider preserving `es6` to make this easier to use?

fi3ework:
Added it back.

chenjiahan (approved):
👍

PR #302, opened 2024-10-17
chore: should check deps of all package.json

Timeless0911:
## Summary

Now `check-dependency-version-consistency` only works in `packages/*` folder, we need to cover all of the packages folder.

Since both `changeset` and `check-dependency-version-consistency` will read `workspaces` field in `package.json`, this PR have done such thing:

- Add more packages glob pattern to `workspaces.packages`
- Config `privatePackages` field of `.changeset/config.json` to disable bump private packages
- Modify the test script of `tests/integration/cli/index.test.ts` since after adding `tests/**` to `workspaces.packages` field, `npx rslib inspect` will report an error

```
npm ERR! could not determine executable to run
```

## Checklist


package.json
```diff
@@ -3,7 +3,11 @@
   "private": true,
   "workspaces": {
     "packages": [
-      "packages/*"
+      "packages/*",
+      "scripts/*",
+      "tests/**",
+      "examples/**",
+      "website"
```
fi3ework, line 10: Why they're removed, how could they resolve @rslib/core after removed.

chenjiahan (commented):
PR Summary pls 👀

package.json
fi3ework: I was wrong, it's not pnpm workspace file.

Timeless0911:
> PR Summary pls 👀

Just editing before

PR #305, opened 2024-10-17
docs: init website structure

Timeless0911:
## Summary

The basic structure of Rslib website.

## Checklist


Timeless0911:
LibConfig should be put into one html instead of multiple html pages, we can continue to write different mdx for each config and import them all in a index page.

fi3ework (dismissed):

fi3ework (approved):

PR #307, opened 2024-10-18
perf(deps): replace fast-glob with tinyglobby

chenjiahan:
## Summary

Tinyglobby is a fast and minimal alternative to globby and fast-glob, it is used by tsup, Vite and Nuxt.

- `tinyglobby` is now in `dependencies` of `@rslib/core` because `rsbuild-plugin-dts` also depends on it. If we prebundle it, there will be two copies of the code.
- Add `absolute: true` because `fast-glob` and `tinyglobby` use different relative formats, `fast-glob` will return `./path/to/file.js`, `tinyglobby` will return `path/to/file.js`. Also the absolute path is more friendly to the Node `fs` APIs.

This PR should reduce 15 sub-dependencies of Rslib:

image

## Related Links

https://github.com/SuperchupuDev/tinyglobby?tab=readme-ov-file#api

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -9,7 +9,7 @@ import {
   mergeRsbuildConfig,
   rspack,
 } from '@rsbuild/core';
-import glob from 'fast-glob';
+import { glob } from 'tinyglobby';
 import {
   DEFAULT_CONFIG_EXTENSIONS,
   DEFAULT_CONFIG_NAME,
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -748,6 +748,7 @@ const composeEntryConfig = async (
     // Turn entries in array into each separate entry.
     const globEntryFiles = await glob(entryFiles, {
       cwd: root,
+      absolute: true,
     });
 
     // Filter the glob resolved entry files based on the allowed extensions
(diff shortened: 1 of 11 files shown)


Timeless0911:
We also prebundled `picocolors` which is used by both `@rslib/core` and `rsbuild-plugin-dts`, is that two copies of `picocolors` breaks? `picocolors` should be a tiny size package.

Timeless0911 (approved):

packages/core/src/config.ts
```diff
@@ -748,6 +748,7 @@ const composeEntryConfig = async (
     // Turn entries in array into each separate entry.
     const globEntryFiles = await glob(entryFiles, {
       cwd: root,
+      absolute: true,
```
fi3ework, line 751: This will make the entry config of Rspack turn to be an absolute path. Although all the test passes. I'm not sure will it introduce any issues? 🤔

chenjiahan: Rspack supports absolute entry path, see https://rspack.dev/config/entry#path-type

I also found out that the `calcLongestCommonPath` method only accepts absolute paths,  which led me to use the `absolute: true` option.

image

fi3ework: Got it!

PR #325, opened 2024-10-23
feat: support set `dts: true` and dts default to bundleless

Timeless0911:
## Summary

- Support `dts: true` to default generate bundleless DTS
- Only generate bundle DTS when `dts.bundle` set to true

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -906,15 +906,25 @@ const composeDtsConfig = async (
   libConfig: LibConfig,
   dtsExtension: string,
 ): Promise<RsbuildConfig> => {
-  const { dts, bundle, output, autoExternal, banner, footer } = libConfig;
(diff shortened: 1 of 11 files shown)


packages/core/src/config.ts
```diff
@@ -906,15 +906,25 @@ const composeDtsConfig = async (
...
+      bundle: false,
+    };
+  }
+
   const { pluginDts } = await import('rsbuild-plugin-dts');
   return {
     plugins: [
       pluginDts({
-        bundle: dts?.bundle ?? bundle,
+        // only set dts.bundle to true can generate bundle DTS
```
fi3ework, line 926:
```suggestion
        // Only setting ⁠dts.bundle to true will generate the bundled d.ts.
```

fi3ework (approved):

PR #326, opened 2024-10-23
feat(create-rslib): add react template and tools(storybook/vitest)

fi3ework:
#### Programmatically generate templates

- The generated template path is in pattern of `[${template}]-[${(tools ? Object.keys(tools) : []).sort()}]-${lang}`.
- Add new tools step (support vitest and storybook for now), `vitest` is applicable for all templates and `storybook` is applicable for UI related template (react).

#### Add React template

- React (and other UI frameworks) is applicable with tool Storybook.

#### Templates tweak

- Adding `typescript` dependency and `dts: true` for JS projects: Emitting dts for JS templates. It's acceptable to write non-TS project sometimes, but dts is recommended to emit and enable partial type check when needed.

#### TODO

- As the templates getting more complex, I wonder does the `tools` step merge into `create-rstack`?
- Is adding TS stuff to JS project also applicable for Rsbuild JS projects?

## Checklist


.nvmrc
```diff
@@ -1 +1 @@
-v18.20.2
```
Timeless0911, line 1: Will this affect the CI environment? I don't think it's necessary to use node version too new here.

packages/create-rslib/fragments/base/react-js/rslib.config.mjs
```diff
@@ -0,0 +1,35 @@
...
+
+export default defineConfig({
+  source: {
+    entry: {
+      index: ['./src/**'],
+    },
+  },
+  lib: [
+    {
+      ...shared,
```
Timeless0911, line 19: no need to use shared here?

packages/create-rslib/fragments/base/react-js/rslib.config.mjs
```diff
@@ -0,0 +1,35 @@
+import { pluginReact } from '@rsbuild/plugin-react';
+import { defineConfig } from '@rslib/core';
+
+const shared = {
+  bundle: false,
+  dts: {
```
Timeless0911, line 6: we can use `dts: true`

packages/create-rslib/fragments/base/react-ts/rslib.config.ts
```diff
@@ -0,0 +1,46 @@
+import { pluginReact } from '@rsbuild/plugin-react';
+import { pluginSass } from '@rsbuild/plugin-sass';
+import { type LibConfig, defineConfig } from '@rslib/core';
+
+const shared: LibConfig = {
+  bundle: false,
+  dts: {
```
Timeless0911, line 7: `dts: true`

packages/create-rslib/fragments/base/react-ts/rslib.config.ts
```diff
@@ -0,0 +1,46 @@
...
+    entry: {
+      index: ['./src/**', '!./src/env.d.ts'],
+    },
+  },
+  lib: [
+    {
+      ...shared,
+      format: 'esm',
+      output: {
+        distPath: {
```
Timeless0911, line 23: The templated package of `node-dual` not set different output folder for esm/cjs, and it should be same with specification in package.json

packages/create-rslib/fragments/tools/vitest-node-ts/package.json
```diff
@@ -0,0 +1,8 @@
+{
+    "scripts": {
```
Timeless0911, line 2: the indent size seem larger

.nvmrc
fi3ework: > Will this affect the CI environment?

No, we specify the node version there.

> I don't think it's necessary to use node version too new here.

For copying files recursively introduced in node 22. It only requires developer's local node version.

packages/create-rslib/fragments/base/react-js/rslib.config.mjs
fi3ework: The PR is still in the WIP stage.

packages/create-rslib/fragments/tools/vitest-node-ts/package.json
fi3ework: we lost format on stage before

packages/create-rslib/fragments/base/react-ts/rslib.config.ts
```diff
@@ -0,0 +1,24 @@
+import { pluginReact } from '@rsbuild/plugin-react';
+import { type LibConfig, defineConfig } from '@rslib/core';
```
chenjiahan, line 2: Unused `LibConfig`

chenjiahan:
> Is adding TS stuff to JS project also applicable for Rsbuild JS projects?

I think developers of JS projects may not want to use TypeScript

Timeless0911:
The PR title can be more precise🥹

fi3ework:
> I think developers of JS projects may not want to use TypeScript

I was wrong, I forget TS can't only check part of the file. Will remove TS check.

chenjiahan (approved):

fi3ework:
The generated template files are numerous and may continue to grow. We can generate these template files during the create-rslib `publish` or `postinstall` stages. However, we need more testing to ensure this approach works. For now, we will keep the template files in the repository to assist with code reviews.

PR #328, opened 2024-10-25
fix: "mf" format output should be minified by default

nyqykk:
## Summary
MF assets are loaded over the network, which means they will not be compressed by the project. Therefore, minifying them is necessary.


packages/core/src/config.ts
```diff
@@ -304,7 +304,8 @@ export function composeMinifyConfig(config: LibConfig): RsbuildConfig {
         jsOptions: {
           minimizerOptions: {
             mangle: false,
-            minify: false,
+            // MF assets are loaded over the network, which means they will not be compressed by the project. Therefore, minifying them is necessary.
```
chenjiahan, line 307: Can we add a test case to check if the MF output is minified?

fi3ework: We could add some config snapshot in unit test for 'MF', just like `esm` and `cjs` does https://github.com/web-infra-dev/rslib/blob/main/packages/core/tests/config.test.ts#L60-L128.

tests/integration/minify/mf/default/rslib.config.ts
```diff
@@ -0,0 +1,30 @@
...
+          singleton: true,
+        },
+        'react-dom': {
+          singleton: true,
+        },
+      },
+    }),
+  ],
+  output: {
+    target: 'node',
```
fi3ework, line 23: `'mf'` also works with node? 🤔

fi3ework (approved):

Issue #330, opened 2024-10-25
[Bug]: Cannot build if import package name include '@'


puiks:
### Version

npmPackages: @rslib/core: 0.0.14

### Details

This is the examples of module federation,  i found that would be error when i build, and i just import the button from @shopify/polaris.

```typescript
import { init, loadRemote } from '@module-federation/enhanced/runtime';
import { Suspense, createElement, lazy } from 'react';
import { CounterButton } from './CounterButton';
import { useCounter } from './useCounter';
import { Button } from '@shopify/polaris'; // what i change

init({
  name: 'rslib_provider',
  remotes: [
    {
      name: 'mf_remote',
      entry: 'http://localhost:3002/mf-manifest.json',
    },
  ],
});

export const Counter: React.FC = () => {
  const { count, increment, decrement } = useCounter();

  return (
    <div>
      <Suspense fallback={<div>loading</div>}>
        {createElement(
          lazy(
            () =>
              loadRemote('mf_remote') as Promise<{
                default: React.FC;
              }>,
          ),
        )}
      </Suspense>
      <Button primary />
      <h2>
        <span id="mf-e2e-lib-title">Counter From Rslib MF Format: </span>
        <span id="mf-e2e-lib-content">{count}</span>
      </h2>
      <CounterButton id="mf-e2e-lib-decrease" onClick={decrement} label="-" />
      <CounterButton id="mf-e2e-lib-increase" onClick={increment} label="+" />
    </div>
  );
};
``` 

And this is the error when building the `mf-react-componet`

![Image](https://github.com/user-attachments/assets/3811a183-2561-4cb3-9b37-5330fbea8320)

I think it might be an invalid identifier including '@' symbol, but how to solve that, could you please help review this？

### Reproduce link

 n

### Reproduce Steps

1. cd examples/module-federation/mf-react-component
2. yarn add @shopify-polaris@6.6.0 
3. add `import {Button} from '@shopify-polaris'` into the **index.tsx**
3. yarn build

nyqykk:
Can you add this pkg into devDependencies? if you want to build this pkg into your assets you should do that because rslib will auto external pkg which is dependencies or peerDependencies.
And in mf format externalType is set to var, it means @xxx is not a standard syntax, so you get this error

nyqykk:
Another solution is to set externalType to global, i have modify this https://github.com/web-infra-dev/rslib/pull/331

puiks:
[...]

PR #331, opened 2024-10-25
fix: use global as mf format external type

nyqykk:
## Summary

vmok uses `global` by default.

When users use a external pkg like `'@xxx'`, this will be transform to 
```
var xxx = @xxx
```
which will cause syntax Error because there's no quotation marks. So set to global to avoid this
## Related Links
close: https://github.com/web-infra-dev/rslib/issues/330

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -656,7 +656,9 @@ const composeExternalsConfig = (
     esm: 'module-import',
     cjs: 'commonjs',
     umd: 'umd',
-    // if use 'var', when users use external package like '@pkg', this will cause syntax error like 'var pkg = @pkg'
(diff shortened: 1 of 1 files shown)


packages/core/src/config.ts
```diff
@@ -656,7 +656,8 @@ const composeExternalsConfig = (
     esm: 'module-import',
     cjs: 'commonjs',
     umd: 'umd',
-    mf: 'var', // same as default value
+    // if use 'var', when users use external package like '@pkg', this will cause syntax error like 'var pkg = @pkg'
```
fi3ework, line 659: More than this, the common usage of browsers should not be `'var'` and not just for the sake of edge considerations.
We could change the explanation for why we choose `global` instead of `window` or other options, perhaps by referencing vmok's practice.

Timeless0911 (approved):

PR #337, opened 2024-10-28
feat: support `tsc --build` to incremental build project references

Timeless0911:
## Summary

Support `tsc --build` to incremental build project references by using TypeScript compiler API.

This feature can be enabled by setting `dts.build: true`, now it only works when `dts.bundle: false`

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -930,6 +930,7 @@ const composeDtsConfig = async (
         // Only setting ⁠dts.bundle to true will generate the bundled d.ts.
         bundle: dts?.bundle ?? false,
         distPath: dts?.distPath ?? output?.distPath?.root ?? './dist',
+        build: dts?.build ?? false,
         abortOnError: dts?.abortOnError ?? true,
         dtsExtension: dts?.autoExtension ? dtsExtension : '.d.ts',
         autoExternal,
--- a/packages/plugin-dts/src/dts.ts
+++ b/packages/plugin-dts/src/dts.ts
@@ -164,7 +164,7 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
       throw Error(
         `Please set ${info}: "${outDir}" in ${color.underline(
           configPath,
-        )} to keep it same as Lib config.`,
+        )} to keep it same as "dts.distPath" or "output.distPath" field in lib config.`,
       );
     }
   }
--- a/packages/plugin-dts/src/dts.ts
+++ b/packages/plugin-dts/src/dts.ts
@@ -1,5 +1,12 @@
 import fs from 'node:fs';
-import { basename, dirname, isAbsolute, join, relative } from 'node:path';
+import {
+  basename,
+  dirname,
+  isAbsolute,
+  join,
+  relative,
+  resolve,
+} from 'node:path';
 import { logger } from '@rsbuild/core';
 import color from 'picocolors';
 import ts from 'typescript';
@@ -110,6 +117,7 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
     tsconfigPath,
     name,
     cwd,
+    build,
     isWatch,
     dtsExtension = '.d.ts',
     autoExternal = true,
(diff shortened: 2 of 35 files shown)


packages/plugin-dts/src/utils.ts
```diff
@@ -95,10 +95,17 @@ export async function addBannerAndFooter(
   const content = await fsP.readFile(file, 'utf-8');
   const code = new MagicString(content);
 
-  banner && code.prepend(`${banner}\n`);
-  footer && code.append(`\n${footer}\n`);
+  if (banner && !content.trimStart().startsWith(banner.trim())) {
+    code.prepend(`${banner}\n`);
+  }
+
+  if (footer && !content.trimEnd().endsWith(footer.trim())) {
```
fi3ework, line 102: When will the code end with the footer while still trying to append it again?

packages/plugin-dts/src/dts.ts
```diff
@@ -133,6 +142,33 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
...
+          normalize(resolve(dirname(configPath), outDir)))
+    ) {
+      const info =
+        rawCompilerOptions.outDir && !rawCompilerOptions.declarationDir
+          ? 'outDir'
+          : 'declarationDir';
+      throw Error(
+        `Please set ${info}: "${outDir}" in ${color.underline(
+          configPath,
+        )} to keep it same as Lib config.`,
```
fi3ework, line 167: Indicate which config to align with.

packages/core/src/config.ts
```diff
@@ -930,6 +930,7 @@ const composeDtsConfig = async (
         // Only setting ⁠dts.bundle to true will generate the bundled d.ts.
         bundle: dts?.bundle ?? false,
         distPath: dts?.distPath ?? output?.distPath?.root ?? './dist',
+        build: dts?.build ?? false,
```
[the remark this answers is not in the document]
Timeless0911: Sound not very good since the js output is not emitted by tsc, so the output of dependencies of the project only contains DTS. This option should be enabled by user manualy to align with the Modern.js Module. The former version of Modern.js Module has enabled it by default when composite is true, but it causes a lot of isses so in later version, a configuration is added

packages/plugin-dts/src/dts.ts
```diff
@@ -133,6 +142,33 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
     ? distPath
     : rawCompilerOptions.declarationDir || './dist';
 
+  if (build) {
+    // do not allow to use bundle DTS when 'build: true' since temp declarationDir should be set by user in tsconfig
+    if (bundle) {
+      throw Error(`Can not set "dts.bundle: true" when "dts.build = true"`);
```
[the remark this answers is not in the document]
Timeless0911: below code still need judelgement of whether build is enabled

packages/plugin-dts/src/dts.ts
Timeless0911: $outdir is the config value that should be configured, users just need to copy it from the terminal output.

packages/plugin-dts/src/utils.ts
Timeless0911: In watch mode of incremental build, only save code with no change will cause an incremental dts emit, which will trigger this fuction of writing banner and footer. It is different from watch mode of no incremental build...

Timeless0911: In watch mode of incremental build, only save code with no change will cause an incremental dts emit, which will trigger this fuction of writing banner and footer. It is different from watch mode of no incremental build...

packages/plugin-dts/src/dts.ts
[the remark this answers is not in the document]
fi3ework: I mean the error message. Mixed `a: ture` and `b = ture`.

packages/plugin-dts/src/dts.ts
fi3ework: Yeah, I mean we should let user to know the detailed field.

Timeless0911: updated

fi3ework (approved):

PR #346, opened 2024-10-30
fix(create-rslib): preserve @rslib/core as workspace

fi3ework:
## Summary

- leave `workspace:*` to the final step.
- bump testing library

## Checklist


Timeless0911:
What is it mean? The template will be published?

fi3ework:
> What is it mean? The template will be published?

No, previously, the version of `@rslib/core` in template is a fixed version like `^0.0.15` which is wrong. I should be `workspace:*` as before. `workspace:*` will be replaced to the correct vesion in runtime.

fi3ework:
> Where is this logic?

Inside create-rstack.

Timeless0911 (approved):

PR #355, opened 2024-10-31
feat: support `raw: false` option in `banner` and `footer`

eunbae0:
## Summary

- Support `raw: false` option in `banner` and `footer`
- The `raw` option is set to `true` by default to maintain compatibility with existing code.

## Checklist


Timeless0911:
Thank you for your PR. But the `raw` option is at the same level with `js/css/dts`, which is not very reasonable when different output file of different types require different configurations.

That's why we only expose string types for user. We do not want to make this config to be too complicated, so in PR of which this feature been imported, there is a description:

#175

> Notice:
	1. only string type is supported and will not be wrapped in a comment.
	2. The default stage is PROCESS_ASSETS_STAGE_OPTIMIZE_SIZE + 1to prevent avoid influence of minimizers
	3. If you have further demand on banner and footer, you can directly use bannerPlugin.

We will explain this in our docs in the future. 

For raw config, users can write string that contain comment or not. And for other config of banner plugin, achievement of DTS is different from js and css and we do not need to encapsule them.

 What's your opinion?

eunbae0:
> Thank you for your PR. But the `raw` option is at the same level with `js/css/dts`, which is not very reasonable when different output file of different types require different configurations.
> 
> That's why we only expose string types for user. We do not want to make this config to be too complicated, so in PR of which this feature been imported, there is a description:
> 
> #175
> 
> > Notice:
> > 1. only string type is supported and will not be wrapped in a comment.
> > 2. The default stage is PROCESS_ASSETS_STAGE_OPTIMIZE_SIZE + 1to prevent avoid influence of minimizers
> > 3. If you have further demand on banner and footer, you can directly use bannerPlugin.
> 
> We will explain this in our docs in the future.
> 
> For raw config, users can write string that contain comment or not. And for other config of banner plugin, achievement of DTS is different from js and css and we do not need to encapsule them.
> 
> What's your opinion?

Hello, thank you for reviewing the PR.

First of all, the purpose of providing the raw option is to allow for the extensibility of overriding options in the [bannerPlugin](https://rspack.dev/plugins/webpack/banner-plugin#options). However, since DTS is not implemented as part of the bannerPlugin, it needs to be implemented directly, and I believe there are no other meaningful options to manipulate besides the raw option.

As you mentioned, to avoid making the configuration too complex, I have implemented the raw option at the same level as js/css/dts. However, if it is clearly stated in the docs that the banner is not wrapped in a comment by default, I think we can skip supporting this `raw: false` feature.

Therefore, it would be fine to close this [PR](https://github.com/web-infra-dev/rslib/pull/355).

P.S. I would like to know the plans for the entire documentation work and whether I can contribute by writing the documentation myself.

fi3ework:
> P.S. I would like to know the plans for the entire documentation work and whether I can contribute by writing the documentation myself.

First, thank you for your contributions. We started writing basic documentation for Rslib this week and aim to complete at least a foundational version of each article within two weeks. Polishing the documentation will be a long journey.

For contributors who are new to Rslib and want to help with documentation, we recommend contributing to the reference page for the config, such as https://lib.rsbuild.dev/config/lib/banner. These pages are more focused, have clearer content, and cover a smaller range of source code. The Rslib documentation will follow the style of Rsbuild, which you can reference https://rsbuild.dev/config/dev/asset-prefix.

You can submit a draft PR when the document is nearly complete to get more feedback from the maintainer in the early stages.

Thanks for your contribution.

PR #367, opened 2024-11-04
chore: export composeCreateRsbuildConfig as utils

nyqykk:
## Checklist

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -3,7 +3,7 @@ export { runCli } from './cli/commands';
 export {
   defineConfig,
   loadConfig,
-  composeCreateRsbuildConfig as _composeCreateRsbuildConfig,
+  composeCreateRsbuildConfig as unstable_composeCreateRsbuildConfig,
 } from './config';
 export { build } from './build';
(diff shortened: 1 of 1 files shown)


chenjiahan (commented):
We should be careful about exporting public APIs, and all public APIs should be documented later.

For private APIs, use prefixes like `_` or `_internal`.

packages/core/src/index.ts
```diff
@@ -1,6 +1,10 @@
 export { prepareCli } from './cli/prepare';
 export { runCli } from './cli/commands';
-export { defineConfig, loadConfig } from './config';
+export {
+  defineConfig,
+  loadConfig,
+  composeCreateRsbuildConfig as _composeCreateRsbuildConfig,
```
fi3ework, line 6:
```suggestion
  composeCreateRsbuildConfig as unstable_composeCreateRsbuildConfig,
```

I'd prefer prefix with `unstable_` just like React (https://github.com/search?q=repo%3Afacebook%2Freact%20unstable&type=code). Which can warn user in a more clear way.

fi3ework (approved):

PR #398, opened 2024-11-11
feat!: target default to `node`

Timeless0911:
## Summary

**Breaking change**

Set `output.target` default to `node` to align with Tsup and Modern.js Module, which is different from Rsbuild default.

## Checklist


chenjiahan (commented):
We can now remove this chapter:

https://lib.rsbuild.dev/guide/migration/tsup#platform

packages/core/src/config.ts
```diff
@@ -941,7 +941,7 @@ const composeDtsConfig = async (
 };
 
 const composeTargetConfig = (
-  target: RsbuildConfigOutputTarget = 'web',
+  target: RsbuildConfigOutputTarget = 'node',
 ): RsbuildConfig => {
```
fi3ework, line 945: Could return in shape of `: { target: RsbuildConfigOutputTarget, rsbuildConfig: RsbuildConfig }`, so that we could use the unified resolved target. The duplicated fallback in line 717 can be removed then.

Timeless0911: great

fi3ework (approved):

Issue #391, opened 2024-11-09
[Bug]: ReferenceError when using wildcard import


chenjiahan:
### Version

```sh
0.0.16
```

### Details

Input:

```js
import * as nodeLibs from './libs.js';
```

Runtime error:

![Image](https://github.com/user-attachments/assets/fe0379fb-64b0-40f1-b0e3-7032baec3def)

### Reproduce link

https://github.com/rspack-contrib/rsbuild-plugin-node-polyfill/tree/rslib_repro_1109

### Reproduce Steps

1. clone the repo
2. git co rslib_repro_1109
3. pnpm i
4. pnpm run test

Issue #381, opened 2024-11-07
Preserving directive in output


fi3ework:
If I want to use rslib to create a component library which supports server components, how can I preserve the `"use client"` directive at the top of the file? 

I have a pretty basic config:
```
import { pluginReact } from '@rsbuild/plugin-react';
import { defineConfig } from '@rslib/core';

export default defineConfig({
  source: {
    entry: {
      index: ['./src/**'],
    },
  },
  lib: [
    {
      bundle: false,
      dts: true,
      format: 'esm',
    },
  ],
  plugins: [
    pluginReact({
      swcReactOptions: {
        runtime: 'classic',
      },
    }),
  ],
});
```

But my output for a simple client components winds up looking like this:
```
import * as __WEBPACK_EXTERNAL_MODULE_react__ from "react";
'use client';
function Counter() {
    const [count, setCount] = __WEBPACK_EXTERNAL_MODULE_react__["default"].useState(0);
    return /*#__PURE__*/ __WEBPACK_EXTERNAL_MODULE_react__["default"].createElement("div", {
        style: {
            display: 'flex',
            gap: 8,
            padding: 8
        }
    }, /*#__PURE__*/ __WEBPACK_EXTERNAL_MODULE_react__["default"].createElement("p", null, "Count: ", count), /*#__PURE__*/ __WEBPACK_EXTERNAL_MODULE_react__["default"].createElement("button", {
        onClick: ()=>setCount((c)=>c + 1)
    }, "Increment"));
}
export { Counter as default };
```

If `"use client";` isn't the first line in the component definition, it's an invalid client component. If I manually move the directive to the top of the file before packing & installing in my consuming app, it works fine. Is there some config options or plugins that exist today to help solve this, or is this currently unsupported?

_Originally posted by @rdenman in https://github.com/web-infra-dev/rslib/discussions/379_

fi3ework:
Currently, there's not a precise way to preserve the directive. Using `banner (https://lib.rsbuild.dev/config/lib/banner) may works but it's will add banner to all files.

This issue is similar to https://github.com/web-infra-dev/rslib/issues/166. `InitFragment` will force inserted the top of the chunk regardless the directive or shebang.

Issue #166, opened 2024-09-02
[Bug]: Rslib build breaks shebang line in bin file


chenjiahan:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.72 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 128.0.6613.86
    Safari: 17.6
  npmPackages:
    @rslib/core: 0.0.4 => 0.0.4
```

### Details

Rslib build breaks shebang line in bin file:

- Input:

![image](https://github.com/user-attachments/assets/deb4aeb4-efc5-489c-b4b6-f42017a8350b)

- Output:

![image](https://github.com/user-attachments/assets/8f66d165-5b81-47f6-b90e-a2373cf3c6fe)

- Failed to run:

![image](https://github.com/user-attachments/assets/9562256e-9dfa-424f-86c2-a178644b05dc)

### Reproduce link

https://github.com/web-infra-dev/rsbuild/tree/8a2a850aad336b726e50b2b1bb203d8c72c1c48a/packages/create-rsbuild

### Reproduce Steps

1. pnpm i
2. pnpm build

fi3ework:
- webpack / Rspack will adding `//` to strip the shebang, see https://github.com/webpack/webpack/pull/13961.
- webpack / Rspack will not hoist shebang to the top of the bundle intentionally.
- esbuild already handles chmod internally https://github.com/evanw/esbuild/commit/b808c0d905551dde55e2485eb18d488b12e4ba94.

Shortly we could use https://github.com/shuangwhywhy/webpack-shebang-plugin. In longterm, we could considering making this a built-in plugin to webpack/Rspack.

chenjiahan:
Yes, we can add a built-in plugin to webpack and Rspack when the `target` is `node`.

PR #399, opened 2024-11-11
feat: add EntryChunkPlugin to handle shebang and shims

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/391.
Fix https://github.com/web-infra-dev/rslib/issues/381.
Fix https://github.com/web-infra-dev/rslib/issues/166.
Fix https://github.com/web-infra-dev/rslib/issues/425.

### Overview

- Change the implementation of shim for `import.meta.url`. Previously, we can't use BannerPlugin to insert it into the top of the bundle as there might be a existing `"use strict"`, inject to the end also does not help as there will be a TDZ. In thi PR, using `processAssets` hook to insert the shim code to the correct position while preserve the leading `"use strict"`.
- Preserve the leading shebang directive for chunk entry modules, the implementation refers https://github.com/shuangwhywhy/webpack-shebang-plugin.
- Preserve the leading [React directive](https://react.dev/reference/rsc/directives) for chunk entry modules.

## Checklist


packages/plugin-dts/src/dts.ts
```diff
@@ -189,14 +189,12 @@ export async function generateDts(data: DtsGenOptions): Promise<void> {
...
-    entry = join(
-      declarationDir!,
-      relativePath,
-      basename(entrySourcePath),
-    ).replace(
-      /\.(js|mjs|jsx|ts|mts|tsx|cjs|cts|cjsx|ctsx|mjsx|mtsx)$/,
-      '.d.ts',
-    );
+    entry = join(declarationDir!, relativePath, basename(entrySourcePath))
+      .replace(/\?.*$/, '')
```
Timeless0911, line 193: leave a comment to specify to query syntax

tests/scripts/shared.ts
```diff
@@ -1,6 +1,6 @@
 import assert from 'node:assert';
 import fs from 'node:fs';
-import { dirname, join, normalize } from 'node:path';
+import path, { dirname, join, normalize } from 'node:path';
```
Timeless0911, line 3: why not use `{ basename }`

packages/core/src/config.ts
```diff
@@ -1100,6 +1123,10 @@ async function composeLibRsbuildConfig(config: LibConfig, configPath: string) {
     cssModulesAuto,
   );
   const cssConfig = composeCssConfig(lcp, config.bundle);
+  const entryChunkConfig = composeEntryChunkConfig({
```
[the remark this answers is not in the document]
fi3ework: Tweaked the return value, do not need to use `format` alongside now.

tests/scripts/shared.ts
fi3ework: I was intended to avoid the name conflict.

packages/core/src/config.ts
```diff
@@ -600,10 +600,14 @@ const composeFormatConfig = ({
   }
 };
 
+type DeepRequired<T> = Required<{
```
Timeless0911, line 603: export in types or utils folder

packages/core/src/config.ts
```diff
@@ -1020,7 +1041,7 @@ const composeTargetConfig = (
 const composeExternalHelpersConfig = (
   externalHelpers: boolean,
   pkgJson?: PkgJson,
-): RsbuildConfig => {
+) => {
```
Timeless0911, line 1044: why isolated declaration not throw an error

fi3ework: Cause it's not exported?

Timeless0911 (approved):

PR #412, opened 2024-11-12
fix: target of mf format should be web

nyqykk:
## Summary
Target of mf format should be web, or will create break change.
Users should not specifically config target to 'web' in their projects

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -949,11 +949,11 @@ const composeTargetConfig = (
   config: RsbuildConfig;
   target: RsbuildConfigOutputTarget;
 } => {
-  let recommendedTarget = target;
-  if (!recommendedTarget) {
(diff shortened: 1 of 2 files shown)


packages/core/src/config.ts
```diff
@@ -941,12 +943,17 @@ const composeDtsConfig = async (
...
 const composeTargetConfig = (
-  target: RsbuildConfigOutputTarget = 'node',
+  target: RsbuildConfigOutputTarget,
+  format: Format,
 ): {
   config: RsbuildConfig;
   target: RsbuildConfigOutputTarget;
 } => {
-  switch (target) {
+  let recommendedTarget = target;
```
fi3ework, line 952: The name is weird, prefer to use `defaultTarget`.

fi3ework (approved):

Issue #404, opened 2024-11-11
[Bug]: Failed to generate DTS when running build twice


chenjiahan:
### Version

```sh
Rslib 0.0.17
```

### Details

- Run Rslib twice: failed

![Image](https://github.com/user-attachments/assets/913c4c35-835f-4cb5-82a6-197c0f9758fe)

- Run tsc twice: succeed

![Image](https://github.com/user-attachments/assets/3bead099-dff1-4974-a401-9dda2abbc957)

### Reproduce link

https://github.com/web-infra-dev/rsbuild/tree/rslib_repro_types_1111

### Reproduce Steps

1. pnpm i
2. cd packages/core
3. pnpm run build && pnpm run build

Timeless0911:
Since `composite: true` is set in `tsconfig.json`, the `incremental` will be default to `true`, thus creating a series of `.tsbuildinfo` files the same folder as your compilation output.

When we use `npx tsc`, it will generate `.tsbuildinfo` file. And for Rslib, we should use `dts.build: true` to align with this incremental build behaviour and this issue can be resolved.

For other Rsbuild plugins in Rsbuild repo, maybe `dts.build: true` should be enabled to keep the same behaviour with tsc. Since these plugins does not specify a special `declarationDir` that just keep the same value with the `outDir`, so this error does not appear.

Set `dts.build: true` will also generate types and buildinfo files for other projects in `references` which should be noted.

Modern.js Module just use bin file of `tsc`, so it do not need to do anything else. Rslib uses Typescript compiler API, and for composite project that have a special `declarationDir`, `dts.build: true` should be set.

Timeless0911:
The `dts.build` options is designed to align with `--build` param for the tsc command. But for some special composite projects like `@rsbuild/core` that do not have `references`, it still need to enable `dts.build` to make it work like tsc normally.

What can we do for some default values of `dts.build`?

option 1. When `composite` is true, `dts.build` default to true. And in order to avoid buildinfo issue, we should automatically delete buildinfo file when `composite` is true too.

[...]

PR #416, opened 2024-11-12
feat: support incremental build when composite is true

Timeless0911:
## Summary

Our current logic using the [Typescript compiler API](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#writing-an-incremental-program-watcher) is:

### build mode

1. `dts.build: false`: [basic compiler](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#a-minimal-compiler) with `createCompilerHost` and `createProgram`
2. `dts.build: true`: incremental build with project references, using `createSolutionBuilderHost` and `createSolutionBuilder`

### watch mode

1. `dts.build: false`: [incremental program watcher](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#writing-an-incremental-program-watcher) with `createWatchCompilerHost` and `createWatchProgram`
2. `dts.build: true`: incremental build with project references, using `createSolutionBuilderWithWatchHost` and `createSolutionBuilderWithWatch`

In this PR, we add `createIncrementalCompilerHost` and `createIncrementalProgram` in build mode to support incremental and deal with projects that with `composite: true`(`incremental: true`) and do not to build references dependencies. The watch mode API originally support incremental, so we do not to deal with it in this scenerio.

Our logic is beyond what [fork-ts-checker-webpack-plugin](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin) do

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/0fab463b21c6edc4d94834568a3f440241d57887/src/typescript/worker/get-issues-worker.ts#L66-L73

For compiler API that not use `SolutionBuilder`, it seems the buildinfo not actually takes effect, it just generate buildinfo files. And for tsbuildinfo, due to various issues caused by buildinfo such as remove distPath and monorepo sceniro, we decide to delete buildinfo before generating dts in the next PR which is aligned with Modern.js Module. And our dts processing behaviour such as redirect alias, modify dts extensions and adding banner/footer are also not compatitle with ts cache.

For users who want to keep most align with the default behaviour with `tsc` in incremetal scenerio, they can enable `dts.build: true` but I think there's no need to do this if they do not want to build dependencies.

Hopefully there won’t be a day where we switch to using tsc bin files directly, and the ts compiler api will be easier to use with more extensive documentation and able to align related behaviors with tsc more easily.

## Related Links

close: #404 

## Checklist

### DIFF
--- a/packages/plugin-dts/src/tsc.ts
+++ b/packages/plugin-dts/src/tsc.ts
@@ -260,7 +260,7 @@ export async function emitDts(
       `DTS generated in ${getTimeCost(start)} ${color.gray(`(${name})`)}`,
     );
   } else {
-    // watch mode
+    // watch mode, can also deal with incremental build
     if (!build) {
       const host = ts.createWatchCompilerHost(
         configPath,
@@ -273,7 +273,7 @@ export async function emitDts(
 
       ts.createWatchProgram(host);
     } else {
-      // incremental build with build project references
+      // incremental build watcher with build project references
       const host = ts.createSolutionBuilderWithWatchHost(
         system,
         createProgram,
--- a/packages/plugin-dts/src/tsc.ts
+++ b/packages/plugin-dts/src/tsc.ts
@@ -18,6 +18,43 @@ export type EmitDtsOptions = {
   footer?: string;
 };
 
+async function handleDiagnosticsAndProcessFiles(
+  diagnostics: readonly ts.Diagnostic[],
+  configPath: string,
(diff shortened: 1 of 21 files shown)


Timeless0911:
Damn! It's really hard to align buildinfo behaviour with tsc command line

fi3ework (approved):

PR #419, opened 2024-11-13
fix: checkMFPlugin function support array plugins

2heal1:
## Summary
plugins may pass array type like this: 

```ts
{
      format: 'mf',
      plugins: [
        [
          { name: 'rsbuild:module-federation-enhanced', setup: () => {} },
          { name: 'plugin-foo', setup: () => {} },
        ],
      ],
}
```
## Checklist

- [X] Tests updated (or not required).

### DIFF
--- a/.changeset/silver-pumas-tap.md
+++ b/.changeset/silver-pumas-tap.md
@@ -1,0 +1,5 @@
+---
+'@rslib/core': patch
+---
+
+fix: checkMFPlugin function support array plugins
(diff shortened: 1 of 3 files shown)


.changeset/silver-pumas-tap.md
```diff
@@ -0,0 +1,5 @@
+---
```
nyqykk, line 1: you can remove changeset file, Rslib release do not need this

Timeless0911 (approved):

PR #424, opened 2024-11-14
perf: enable compress minify option for node target

chenjiahan:
## Summary

For the Node target, keep output bundles as small as possible while retaining the necessary information for debugging, as Node outputs usually executed directly at runtime rather than being built again.

See https://github.com/web-infra-dev/rsbuild/pull/3966 for more.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -3,6 +3,7 @@ import path, { dirname, extname, isAbsolute, join } from 'node:path';
 import {
   type EnvironmentConfig,
   type RsbuildConfig,
+  type RsbuildTarget,
   defineConfig as defineRsbuildConfig,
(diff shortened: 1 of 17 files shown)


chenjiahan:
I found that this optimization breaks a lot of test snapshots, some of the changes are unexpected. For example, the code readability may be significantly worse.

I'm now tending to make this optimization opt-in. If users want a smaller Node.js bundle, they can manually set the minify configuration.

Timeless0911:
> I'm now tending to make this optimization opt-in. If users want a smaller Node.js bundle, they can manually set the minify configuration.

Do we need to encapsulate a configuration, or we can just document how to configure it if a smaller bundle is needed?

chenjiahan:
I think adding documentation is good enough for this

PR #432, opened 2024-11-16
feat(create-rslib): add prepublishOnly script for templates

wangjue666:
## Summary
Automatically execute the `npm run build` command before executing the `npm publish` command

## Checklist

### DIFF
--- a/packages/create-rslib/template-[node-dual]-[]-js/package.json
+++ b/packages/create-rslib/template-[node-dual]-[]-js/package.json
@@ -15,7 +15,8 @@
   ],
   "scripts": {
     "build": "rslib build",
-    "dev": "rslib build --watch"
+    "dev": "rslib build --watch",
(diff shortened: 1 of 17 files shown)


fi3ework:
Thanks for contribution. While we're trying to make `create-rslib` a minimal scaffold to create Rslib project. `prepublishOnly` is not must-have script if the user already ran `npm run build` in the release workflow ahead. Since the workflow might varies, we tend to keep it simple and leave the opt-in command to userland.

wangjue666:
> Thanks for contribution. While we're trying to make `create-rslib` a minimal scaffold to create Rslib project. `prepublishOnly` is not must-have script if the user already ran `npm run build` in the release workflow ahead. Since the workflow might varies, we tend to keep it simple and leave the opt-in command to userland.

Appreciate your insights!

PR #438, opened 2024-11-17
feat!: use 'commonjs-import' as CJS external type by default

fi3ework:
## Summary

It's hard to cover all cases of whether it's under node `13.2.0` without calling Browserslist API. Since Node 13.2 has reach EOL for years (https://endoflife.date/nodejs). I think it's fine to set `commonjs-import` as the default external type for Node.js and demonstrate the differences between the externals type on library build scenario next.

Also bump Rsbuild to `~1.1.4`

## Checklist

### DIFF
--- a/examples/module-federation/mf-host/package.json
+++ b/examples/module-federation/mf-host/package.json
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "@module-federation/rsbuild-plugin": "^0.7.5",
-    "@rsbuild/core": "~1.1.3",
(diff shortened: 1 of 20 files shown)


fi3ework:
TODO: don't merge this until Rsbuild has relaid on `@rspack/core` 1.1.2. It strictly requires at least `1.1.3` version of Rsbuild (assume that Rsbuild will bump `@rspack/core` on `1.1.3`). We need to change the dependent Rsbuild version to `^1.1.3` alongside.

Timeless0911 (approved):

Issue #445, opened 2024-11-18
Add website links in JSDoc


Timeless0911:
Cleaned. We need to add link for all field in another PR, like TypeScript.

_Originally posted by @fi3ework in https://github.com/web-infra-dev/rslib/pull/436#discussion_r1845935926_

PR #447, opened 2024-11-18
fix: add tsdoc for config options

fi3ework:
## Summary

Close https://github.com/web-infra-dev/rslib/issues/445.

Prefer to use [TSDoc](https://tsdoc.org/) as it's doc comments for TypeScript.

## Checklist

### DIFF
--- a/packages/core/src/types/config/index.ts
+++ b/packages/core/src/types/config/index.ts
@@ -80,59 +80,67 @@ export type Redirect = {
 export interface LibConfig extends RsbuildConfig {
   /**
    * The unique identifier of the library.
-   * @default undefined
+   * @defaultValue `undefined`
    */
   id?: string;
   /**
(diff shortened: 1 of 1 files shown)


packages/core/src/types/config/index.ts
```diff
@@ -80,59 +80,67 @@ export type Redirect = {
 export interface LibConfig extends RsbuildConfig {
   /**
    * The unique identifier of the library.
-   * @default undefined
+   * @defaultValue `undefined`
```
Timeless0911, line 83: link

Timeless0911, line 83:
```suggestion
   * @defaultValue `undefined`
   * @see {@link https://lib.rsbuild.dev/config/lib/id}
```

Timeless0911 (approved):

Issue #361, opened 2024-11-02
When you define an external rslib config file with the `-c` parameter, it can't find the package.json anymore.


fi3ework:
When you define an external rslib config file with the `-c` parameter, it can't find the package.json anymore. For example:

```
rslib build -c ./etc/rsbuild/config.ts
```

It will complain that it can't find the package.json from the root folder.

_Originally posted by @binarykitchen in https://github.com/web-infra-dev/rslib/discussions/359_

Timeless0911:
Maybe we should use `process.cwd()` as a fallback root path to load `package.json`.

binarykitchen:
... and unit tests. What if rslib is called from a subdirectory where no config nor package.json exist?

Timeless0911:
> ... and unit tests. What if rslib is called from a subdirectory where no config nor package.json exist?

You mean we need to find the closest config file / package.json from bottom to up? If so, `-c` parameter is not enough, and we also need a `--root` parameter to specify where `package.json` folder is.

binarykitchen:
Well, too many parameters overcomplicate, which isn't good. Keep it simple. The `-c` parameter adds extra complexity.

rslib, rsbuild and rspack are supposed to be lightweight, to be a lot simpler than webpack. So, I wouldn't add more parameters but remove `-c` as well. Try to keep it simple.

Timeless0911:
`-c` parameter should be preserved since config file can be named like `rslib.build.config.ts` just like `tsconfig.build.json`, and I totally agree to not add any other parameters to make things complexed. 

I think we can optimize the logic of finding `package.json` to handle most of situations but not all.

binarykitchen:
@Timeless0911 Ok, good. Invite me for the code review, happy to throw in some ideas maybe.

Issue #356, opened 2024-10-31
[Bug]: Error when calling build via JavaScript API


canyuegongzi:
### Version

"@rslib/core": "^0.0.15",

### Details

错误：
```
× Failed to build.
error   TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

示例代码：
```
import {LibConfig, RslibConfig, build as rsLibBuild} from "@rslib/core";
async function buildFun(): Promise<any> {
  console.log("build")
  const defaultConfig: RslibConfig = {
    lib: [
      {
        format: 'esm',
        source: {
          entry: {
            index: [path.resolve(process.cwd(), './src/**')],
          }
        },
        output: {
          distPath: {
            root: path.resolve(process.cwd(), './lib/esm'),
          }
        },
      }
    ],
    plugins: [
      pluginReact({
        swcReactOptions: {
          runtime: 'classic',
        },
      }),
      pluginSass(),
    ],
  }
  console.log("构建配置")
  console.log(JSON.stringify(defaultConfig, null, 2))
  await rsLibBuild(defaultConfig);
}
```

### Reproduce link

https://github.com/canyuegongzi/rslib-example-bug

### Reproduce Steps

1: npm install
2: npm run build:script

Timeless0911:
Thanks, the js api of Rslib are not public yet. We will track it and write clear documentation when it is available.

canyuegongzi:
> Thanks, the js api of Rslib are not public yet. We will track it and write clear documentation when it is available.

ok

PR #453, opened 2024-11-19
feat: add `--root` cli option and make basic js api work

Timeless0911:
## Summary

- Add `--root` CLI param as base dir to load Rslib config file and .env file
- Use `config.root` which is default to `process.cwd()` as base dir to load `package.json` and `tsconfig.json`

close: #356 
close: #361 

## Related Links

#456 

## Checklist


Timeless0911:
@binarykitchen I think this PR can solve your issue.

packages/core/src/cli/build.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { RslibConfig } from '../types/config';
+import { getAbsolutePath } from '../utils/helper';
+import type { BuildOptions } from './commands';
+
+export async function build(
+  config: RslibConfig,
+  options: Pick<BuildOptions, 'root' | 'lib' | 'watch'> = {},
+): Promise<RsbuildInstance> {
+  const cwd = process.cwd();
+  options.root = options.root ? getAbsolutePath(cwd, options.root) : cwd;
```
fi3ework, line 12: `options.root ` has is calculated in `packages/core/src/cli/init.ts#10` and passed to hereby. Do we need to re-calculate it again? (so does mf).

packages/core/src/cli/init.ts
```diff
@@ -0,0 +1,19 @@
+import { loadConfig } from '../config';
+import type { RslibConfig } from '../types';
+import { getAbsolutePath } from '../utils/helper';
+import type { CommonOptions } from './commands';
+
+export async function getRslibConfig(
+  options: CommonOptions,
+): Promise<{ root: string; rslibConfig: RslibConfig }> {
+  const cwd = process.cwd();
+  const root = options.root ? getAbsolutePath(cwd, options.root) : cwd;
```
fi3ework, line 10: IIUC, the path of config file does not take account into the `root` calculation anymore?

packages/core/src/cli/build.ts
Timeless0911: This `export async function build(
  config: RslibConfig,
  options: Pick<BuildOptions, 'root' | 'lib' | 'watch'> = {},
): Promise<RsbuildInstance>` function is aim to be our javascript api, so the absolute root path should be generated since `packages/core/src/cli/init.ts#10` is the cli logic. This file is put into the `cli` folder which may be not very good. Since we are not design the javascript api well, we can keep it here.

packages/core/src/cli/init.ts
Timeless0911: Yes, the `root` is as the base dir to find to config files.

packages/core/src/cli/build.ts
fi3ework: It's better not to modify parameter to avoid potential bug.

binarykitchen: Agree. Anything async is very much prone to bugs. And like @fi3ework said, modifying parameters is a no-go.

Function parameters should remain **immutable**.


_Remember, UI = fn(data)_

packages/core/src/cli/inspect.ts
binarykitchen, line 1: Wow, Webpack doesn't have this :)

Interesting ... just a little feedback, the function name `inspect` what does it truly inspect? If it changes configuration, I'd rename the function maybe?

packages/core/tests/__snapshots__/config.test.ts.snap
```diff
@@ -72,7 +72,7 @@ exports[`Should compose create Rsbuild config correctly > Merge Rsbuild config 1
           "pnpapi",
         ],
         "filename": {
-          "js": "[name].js",
+          "js": "[name].mjs",
```
binarykitchen, line 75: Why the change to mjs?

tests/integration/cli/build.test.ts
binarykitchen, line 1: Thanks for adding unit tests +++++++

packages/core/src/cli/commands.ts
```diff
@@ -88,21 +91,13 @@ export function runCli(): void {
...
-          envMode: options.envMode,
-        });
-        const environments = await composeRsbuildEnvironments(rslibConfig);
-        const rsbuildInstance = await createRsbuild({
-          rsbuildConfig: {
-            environments: pruneEnvironments(environments, options.lib),
-          },
-        });
-        await rsbuildInstance.inspectConfig({
+        const { root, rslibConfig } = await getRslibConfig(options);
```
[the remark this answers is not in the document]
Timeless0911: Environment variables may not be passed in during runtime since user may use js api to start the child process without inherit some env. 

The reason for passing root is to read tsconfig and package.json. I think we can extract the relevant logic to the outside when we doing configuration verification and no configuration file mode in the future.

packages/core/tests/__snapshots__/config.test.ts.snap
Timeless0911: Because of the directory of the test case, package.json was not resolved successfully here just like your issue before, so `autoExtension` was not enabled.

packages/core/src/cli/inspect.ts
Timeless0911: see https://lib.rsbuild.dev/guide/basic/cli#rslib-inspect, it does not changes configuration but print detailed configuration to better help users debug.

packages/core/tests/__snapshots__/config.test.ts.snap
binarykitchen: Ah, looks like this will fix more bugs than just one ...
... do we have a unit test case for this?

Timeless0911: Current tests is enough since it correctly use the root package.json to generate extensions. This snapshot is used to check config merge behaviour, we do not find the issue of not finding package.json in this snapshot before. So this snapshot is just the unit test case for this.
 
And about autoExtension, there already have unit tests.

fi3ework (dismissed):

packages/core/src/cli/commands.ts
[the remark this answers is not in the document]
binarykitchen: > And also we will design our javascript API, so the root can be put into a context to pass around.

Ok, good, thanks for explaining. Do you already have a ticket for this?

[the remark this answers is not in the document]
Timeless0911: We are discussing that we should follow Rsbuild to resolve package.json and tsconfig.json, I think the package.json should be in folder as same as webpack context. https://rsbuild.dev/config/root#root

fi3ework (approved):

fi3ework (approved):

PR #481, opened 2024-11-26
test: avoid potential output conflict

fi3ework:
## Summary

Test files are parallelly, the build and inspect test file will conflict with each other. Not occurred on CI yet, But in theory it will and on my device locally now.

## Checklist


Timeless0911:
I think we can create unique folder for test cases, now it shared one folder to run.

fi3ework:
> I think we can create unique folder for test cases, now it shared one folder to run.

Damn, I'm wondering why I'm not doing this way.

Timeless0911 (approved):

PR #484, opened 2024-11-27
fix: entry query should handle non-string entry

fi3ework:
## Summary

Correctly handle all possible entry types to add entry query. Previously, we can only handle `string` indeed.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -39,6 +39,8 @@ import type {
   LibOnlyConfig,
   PkgJson,
   Redirect,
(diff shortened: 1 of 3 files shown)


Timeless0911:
Can we also support `EntryDescription` in compose bundleless entry?

fi3ework:
> Can we also support `EntryDescription` in compose bundleless entry?

I don't think user need to use `EntryDescription` in bundless mode, even in bundle mode for foreseeable usage. We could talk about that in the future.

Timeless0911 (approved):

PR #487, opened 2024-11-28
fix: the moduleIds should be deterministic if the env is production and format is mf

2heal1:
## Summary

When format is mf , and env is production , the moduleIds should be deterministic for better cache and smaller size. 

And the named moduleIds may cause domain name leaked if the build env path has sensitive infomation

## Checklist

- [X] Tests updated (or not required).
- [X] Documentation updated (or not required).

### DIFF
--- a/.changeset/smooth-parents-pull.md
+++ b/.changeset/smooth-parents-pull.md
@@ -1,0 +1,5 @@
+---
+'@rslib/core': patch
+---
+
+fix: the moduleIds should be deterministic if the env is production and format is mf
--- a/packages/core/src/cli/mf.ts
+++ b/packages/core/src/cli/mf.ts
@@ -41,6 +41,7 @@ function changeEnvToDev(rsbuildConfig: RsbuildConfig) {
       rspack: {
         optimization: {
           nodeEnv: 'development',
+          moduleIds: 'named',
         },
       },
     },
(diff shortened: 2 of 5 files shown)


nyqykk:
Can we add this configuration here instead of another tools function? https://github.com/web-infra-dev/rslib/blob/5c8a73b9f1a77c3e9a2213060700a35733d60c0d/packages/core/src/config.ts#L583

packages/core/src/utils/helper.ts
```diff
@@ -201,4 +201,20 @@ export function checkMFPlugin(config: LibConfig): boolean {
   return added;
 }
 
+export function composeMFPlugin(config: LibConfig): RsbuildConfig {
+  if (config.format !== 'mf' || process.env.NODE_ENV !== 'production') {
```
Timeless0911, line 205: `rslib build` and `rslib build --watch` all uses production mode by default, so the default moduleIds should be `deterministic` by default. Why we do this?

Timeless0911: Got it, so ``process.env.NODE_ENV !== 'production'` is used to distinguish `mf dev`

Timeless0911 (dismissed):

nyqykk (approved):

PR #489, opened 2024-11-28
feat: support watch `rslib.config` to rebuild

9aoy:
## Summary

image

image

## Related Links

closes: https://github.com/web-infra-dev/rslib/issues/138

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #494, opened 2024-11-29
fix: lazy import chokidar

9aoy:
## Summary

https://github.com/web-infra-dev/rslib/pull/489

## Checklist

### DIFF
--- a/packages/core/src/cli/restart.ts
+++ b/packages/core/src/cli/restart.ts
@@ -1,5 +1,4 @@
 import path from 'node:path';
-import chokidar from 'chokidar';
 import { color, debounce, isTTY } from '../utils/helper';
 import { logger } from '../utils/logger';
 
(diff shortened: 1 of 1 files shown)


packages/core/src/cli/restart.ts
```diff
@@ -11,11 +10,12 @@ export async function watchFilesForRestart(
     return;
   }
 
+  const chokidar = await import('chokidar');
+
   const watcher = chokidar.watch(files, {
     ignoreInitial: true,
     // If watching fails due to read permissions, the errors will be suppressed silently.
     ignorePermissionErrors: true,
-    ignored: ['**/node_modules/**', '**/.git/**', '**/.DS_Store/**'],
```
fi3ework, line 18: Why ignored is removed here, I think we need to do glob manually due to https://github.com/paulmillr/chokidar#upgrading.

9aoy: This method is only used to watch config files. Currently, there is no scenario requirement for glob and ignored.

fi3ework (approved):

PR #507, opened 2024-12-02
chore: use `resolve.alias` instead of `source.alias`

Timeless0911:
## Summary

We need to update docs in https://lib.rsbuild.dev/config/rsbuild/source#sourcealias in the future, the Rsbuild docs of `resolve.alias` and `resolve.aliasStrategy` is not ready now.

## Related Links

feat: add resolve.alias and deprecate source.alias by @chenjiahan in https://github.com/web-infra-dev/rsbuild/pull/4098

## Checklist


chenjiahan (approved):
👍

Issue #499, opened 2024-11-29
[Bug]: Target "node" output a static folder with bad exports


patricklafrance:
### Version

```sh
System:
    OS: Windows 11 10.0.22631
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-13700H
    Memory: 8.92 GB / 31.62 GB
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
```

### Details

I am using version `0.1.0` of Rslib.

When my config has `target: "node"`, Rslib seems to output a `static/assets/index.ts` file. My project doesn't include any assets. The content of this `static/assets/index.ts` match the content of the `src/index.ts` file:

```ts
export * from "./defineConfig.ts";
export * from "./exports.ts";
```

Here's my Rslib config:

```ts
export default defineConfig({
    lib: [{
        format: "esm",
        syntax: "esnext",
        bundle: true,
        dts: {
            bundle: true
        }
    }],
    source: {
        entry: {
            index: "./src/index.ts",
            sharedDependenciesResolutionPlugin: "./src/sharedDependenciesResolutionPlugin.ts",
            nonCacheableRemoteEntryPlugin: "./src/nonCacheableRemoteEntryPlugin.ts"
        }
    },
    output: {
        target: "node",
        distPath: {
            root: "./dist"
        },
        cleanDistPath: true,
        minify: false
    },
    tools: {
        htmlPlugin: false
    }
});
```

![Image](https://github.com/user-attachments/assets/ebb256cc-fd50-49d4-960f-4ee60505d092)

![Image](https://github.com/user-attachments/assets/af61d2af-2abc-4b82-8750-2b05e3b60f02)

### Reproduce Steps

- `rslib build`

Timeless0911:
Can you provide a reproduction demo? This behaviour seems a bit strange.

patricklafrance:
Here's the repro: https://github.com/patricklafrance/rslib-issue-499-repro

fi3ework:
We'have inspected it's due to this PR of Rsbuild https://github.com/web-infra-dev/rsbuild/pull/4011, which is not applicable for lib scenario. We'll make a quick fix.

elecmonkey:
Hello. Rslib 1.0 has been officially released, and it includes a breaking change that affects the code mentioned in your issue. You can use `/* rspackIgnore: true */` to skip processing that specific `new URL`.

[...]

PR #509, opened 2024-12-02
fix(esm): disable module.parser.javascript.url

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/499.

Disable `module.parser.javascript.url` as it's rarely used in library build scenario.

## Checklist


Timeless0911:
Why is this case be named `format`

packages/core/src/config.ts
```diff
@@ -612,6 +613,19 @@ const composeFormatConfig = ({
   }
 };
 
+const formatRsbuildPlugin = (): RsbuildPlugin => ({
```
Timeless0911, line 616: Does modify `module.parser.javascript.url` not takes effect?

fi3ework: it will affect less url parsing.

Timeless0911: @SoonIter

Timeless0911: https://github.com/web-infra-dev/rslib/actions/runs/12115926632/job/33775293755?pr=509

Timeless0911 (approved):

PR #516, opened 2024-12-04
fix(create-rslib): set tsconfig jsx to `react-jsx`

chenjiahan:
## Summary

React >= 17 should use `react-jsx` instead of `react`.

image

## Related Links

https://www.typescriptlang.org/tsconfig/#jsx

## Checklist


fi3ework (commented):
This was for compatibility reasons. As we didn't require a minimal `react >= 16.14.0` in peer dep. https://github.com/web-infra-dev/rslib/blob/c3857eb19ab801b8fe96d715349b4c08adc1c35f/examples/react-component-bundle/rslib.config.ts#L32 need to be updated alongside.

I checked some mainstream React libraries:

- antd: classic
- @tanstack/react-query: automatic
- react-router: classic
- mobx-react: classic
- framer-motion: automatic

I believe moving to the new JSX runtime for new libraries is a positive step. This change can help avoid potential migration costs in the future.

chenjiahan:
Agreed, I have updated `rslib.config.*` to match with 'react-jsx`.

fi3ework (approved):

PR #520, opened 2024-12-04
feat: do not preserve nodeEnv when format is umd

Timeless0911:
## Summary

do not preserve nodeEnv when format is umd

## Checklist


chenjiahan (approved):
LGTM 👍

PR #531, opened 2024-12-06
feat: disable autoExternal when format is umd or mf

Timeless0911:
## Summary

 Disable `autoExternal` when `format` is `umd` or `mf` since their outputs will be directly consumed.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -58,6 +58,7 @@ import {
   color,
   getAbsolutePath,
   isEmptyObject,
+  isIntermediateOutputFormat,
   isObject,
   nodeBuiltInModules,
   omit,
@@ -235,13 +236,14 @@ const composeExternalsWarnConfig = (
 };
 
 export const composeAutoExternalConfig = (options: {
-  autoExternal: AutoExternal;
+  format: Format;
+  autoExternal?: AutoExternal;
   pkgJson?: PkgJson;
   userExternals?: NonNullable<RsbuildConfig['output']>['externals'];
 }): RsbuildConfig => {
-  const { autoExternal, pkgJson, userExternals } = options;
+  const { format, autoExternal = true, pkgJson, userExternals } = options;
 
-  if (!autoExternal) {
+  if (autoExternal === false || !isIntermediateOutputFormat(format)) {
     return {};
   }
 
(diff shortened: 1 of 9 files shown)


2heal1 (approved):

packages/core/src/config.ts
```diff
@@ -235,13 +236,14 @@ const composeExternalsWarnConfig = (
...
+  format: Format;
+  autoExternal?: AutoExternal;
   pkgJson?: PkgJson;
   userExternals?: NonNullable<RsbuildConfig['output']>['externals'];
 }): RsbuildConfig => {
-  const { autoExternal, pkgJson, userExternals } = options;
+  const { format, autoExternal = true, pkgJson, userExternals } = options;
 
-  if (!autoExternal) {
+  if (autoExternal === false || !isIntermediateOutputFormat(format)) {
```
fi3ework, line 246: Does this mean `umd` and `mf` could never use `autoExternal` anymore?

fi3ework: What about make the default value to `true` for esm and cjs，and `false` for others. Instead of ban the field as of now.

fi3ework: Still will not work, when format is`"mf"`, it still will directly return. You could use `userExternals` to determine if the config is specified.

fi3ework: Got it man.

Timeless0911: updated the commit, you can review it again.

packages/core/src/config.ts
```diff
@@ -234,14 +235,29 @@ const composeExternalsWarnConfig = (
   };
 };
 
+const getAutoExternalDefaultValue = (
+  format: Format,
+  autoExternal?: AutoExternal,
+): AutoExternal => {
+  return isIntermediateOutputFormat(format)
+    ? (autoExternal ?? true)
+    : (autoExternal ?? false);
```
fi3ework, line 244:
```suggestion
  return autoExternal ?? isIntermediateOutputFormat(format);
```

Timeless0911: Cursor is stupid and so am I.

fi3ework (approved):

Issue #407, opened 2024-11-11
[Bug]: wrong tsconfig.json#paths output when `bundle: false`


carlosmfreitas2409:
### Version
```
Rslib 0.0.16
```

### Details

When using `bundle: false`, the TS path aliases are not being correctly generated in the build.
When using `bundle: true`, it generates correctly, but it is not what I want.

- tsconfig.json
```json
{
  "extends": "@repo/tsconfig/react.json",
  "include": ["src", "stories"],
  "compilerOptions": {
    "paths": {
      "@/*": ["./src/*"]
    }
  }
}
```

- rslib config
![Image](https://github.com/user-attachments/assets/bc5b0da5-a95f-4a2d-a550-063ea5f222b8)

- rslib output
![Image](https://github.com/user-attachments/assets/e3e20b29-90f1-478c-87c9-185244d8dcd9)

- build script
`rslib build`

fi3ework:
Since the request is externalized, it won't follow Rsbuild's default alias behavior(https://rsbuild.dev/guide/advanced/alias). It's better to be aligned.

carlosmfreitas2409:
Hi @fi3ework.

Sorry, I didn't quite understand what you meant.
Are you saying to use `source.alias` and `source.aliasStrategy` in rslib config?
If so, I've tried using them too, but it didn't change anything.

fi3ework:
`source.alias` or `source.aliasStraategy` won't make it work. The alias to an source path (which means it's dynamic after resolved) need `getResolve` function https://github.com/web-infra-dev/rslib/issues/238#issuecomment-2376536224.

Timeless0911:
https://github.com/web-infra-dev/rspack/pull/7348 external function `getResolve` will be supported after this pr merged.

yf-yang:
Sorry, I cannot keep up with your terminology. Does this mean that _when using path aliases and building bundless DTS, the aliases would be replaced with the correct relative paths_ after https://github.com/web-infra-dev/rspack/pull/7348 is done?

Timeless0911:
> Does this mean that when using path aliases and building bundless DTS, the aliases would be replaced with the correct relative paths after https://github.com/web-infra-dev/rspack/pull/7348 is done?

For bundleless DTS, the alias redirect should be another issue, tracking in https://github.com/web-infra-dev/rslib/issues/409

noshower:
[...]

Issue #250, opened 2024-09-30
[Bug]: mainFiles `autoExtensions` bad case `"./components/button/index.js"`


SoonIter:
### Version

```sh
rslib 0.0.8
```

### Details

<img width="400" alt="image" src="https://github.com/user-attachments/assets/877241bc-9c94-4c0b-9dbf-4f7497dfdd0e">

```
.
├── src
│   ├── button
│   │   └── index.tsx
│   └── index.ts
└── tsconfig.json
```

#### input
```ts
// index.ts
import button from './button';
```

#### expected behavior

```javascript
import button from './button/index.mjs';
```

#### actual behavior
```javascript
import button from './button.mjs';
```

### Reproduce link

none

### Reproduce Steps

none

SoonIter:
By the way, in bundless mode, if cannot resolve, error message should be emitted

fi3ework:
Does other support patch `mainFiles` as of now? For bundle mode, support `index` in `mainFields` is fine. In bundleless mode, I feel it's too lose somehow.

fi3ework:
esbuild by default could handle `index` mainFiles, but when the target module is externalized, it will not adding `index` as well.

https://esbuild.github.io/try/#YgAwLjI0LjAALS1idW5kbGUgLS1mb3JtYXQ9ZXNtIC0tZXh0ZXJuYWw6Jy4vYScAZQBlbnRyeS5qcwBpbXBvcnQgeCBmcm9tICcuL2EnCmV4cG9ydCB7IHggfQAAYS9pbmRleC5qcwBleHBvcnQgZGVmYXVsdCA1

If we wish to support this feature, we need to use `getResolve` in the external function to get the resolved path, but currently `getResolve` is not supported.

I think we can temporarily disable `mainFiles` in bundleless mode to avoid this too loose behavior.

Timeless0911:
Same as https://github.com/web-infra-dev/rslib/issues/238, we should check whether getResolve hook in Rspack will be supported in the nearly future.

Timeless0911:
We should also handle when `./folder/index.ts` and `./folder.ts` exists at the same time, there may also be .ts, .tsx, .jsx file existing.

fi3ework:
> We should also handle when ./folder/index.ts and ./folder.ts exists at the same time

We could just follow Rspack's default order.

> there may also be .ts, .tsx, .jsx file existing.

`getResolve` could handle all of these.

joshwooding:
[...]

Issue #238, opened 2024-09-26
[Bug]: Adding js extensions for index resolution


Timeless0911:
### Version

```sh
-
```

## input
```ts
// folder/index.ts
export const folder = 'folder';
```

```ts
// index.ts
import { folder } from './folder';

export const text = folder;
```

## Output

```js
import * as __WEBPACK_EXTERNAL_MODULE__folder_js__ from "./folder.js";
const src_text = __WEBPACK_EXTERNAL_MODULE__folder_js__.folder;
export { src_text as text };
```

https://github.com/web-infra-dev/rslib/blob/600937cebd061459a43d4f3e81bf09c96aee0375/packages/core/src/config.ts#L693-L721

In line 710, we directly add a jsExtension in bundless mode to make bundleless esm outputs work.

https://github.com/web-infra-dev/rslib/blob/600937cebd061459a43d4f3e81bf09c96aee0375/packages/core/src/config.ts#L710

**Modern.js Module force user to add `/index` in bundleless mode.**

### Reproduce link

-

### Reproduce Steps

-

fi3ework:
Feeling like we really need `getResolve` function.

Timeless0911:
`getResolve` hook has been supported in externals.

Issue #251, opened 2024-09-30
[Feature]: in bundless, if cannot resolve, error message should be emitted


SoonIter:
By the way, in bundless mode, if cannot resolve, error message should be emitted

_Originally posted by @SoonIter in https://github.com/web-infra-dev/rslib/issues/250#issuecomment-2382455625_

fi3ework:
> if cannot resolve, error message should be emitted

In this case, it's because the resolved module is externalized, Rspack won't check if it's available anymore.

Yes, we need to ensure that the emitted files could be resolved. In this case, we overlooked the ⁠index mainFields, which makes this a fixable bug for us. Let's see if there are any paths that Rslib could not detect in the future. If so, we can add a post-check.

Timeless0911:
`getResolve` in externals function related.

We can resolve every module when externalized and print a warning if cannot be resolved, but we should benchmark the performance of this.

Timeless0911:
https://github.com/web-infra-dev/rslib/blob/8a9f9e336a84a0c7359d5e7d5ad06e8699ea33e2/packages/core/src/config.ts#L1016-L1023

@SoonIter should we throw error/warning here if we cannot resolve in bundleless mode?

SoonIter:
> https://github.com/web-infra-dev/rslib/blob/8a9f9e336a84a0c7359d5e7d5ad06e8699ea33e2/packages/core/src/config.ts#L1016-L1023
> 
> @SoonIter should we throw error/warning here if we cannot resolve in bundleless mode?

let the bullet in the sky, users would feedback

Timeless0911:
https://github.com/web-infra-dev/rslib/blob/6f11138094bb09cec6e056a930a572f2c8608dd1/packages/core/src/config.ts#L1085-L1093  Open a new issue if you want to feedback about this issue.

PR #535, opened 2024-12-08
feat: use resolver function to resolve module

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/238.
Fix https://github.com/web-infra-dev/rslib/issues/250.
Fix https://github.com/web-infra-dev/rslib/issues/553.
Fix https://github.com/web-infra-dev/rslib/issues/407.
Close https://github.com/web-infra-dev/rslib/issues/548.

As of now, we got `getResolve` function in external callback. Invoke `getResolve` once and cache the resolver to use afterward for performance concern.

## Checklist


Timeless0911:
#251
#407

Also should be considered in this PR.

packages/core/src/config.ts
```diff
@@ -955,64 +959,113 @@ const composeEntryConfig = async (
...
   bundle: boolean,
-): RsbuildConfig => {
-  if (bundle) return {};
+): {
+  config: RsbuildConfig;
+  resolvedJsRedirect?: DeepRequired<JsRedirect>;
+} => {
+  if (bundle) return { config: {} };
+
+  const doesRedirectStyle = redirect.style ?? true;
```
Timeless0911, line 973: `isStyleRedirect`

packages/core/src/config.ts
```diff
@@ -955,64 +959,113 @@ const composeEntryConfig = async (
...
+  const jsRedirectPath = redirect.js?.path ?? true;
+  const jsRedirectExtension = redirect.js?.extension ?? true;
 
-  const isStyleRedirect = redirect.style ?? true;
+  type Resolver = GetAsyncFunctionFromUnion<
```
Timeless0911, line 977: put it into `types` folder.

packages/core/src/config.ts
```diff
@@ -955,64 +959,113 @@ const composeEntryConfig = async (
...
+                return cssExternal;
+              }
+
+              if (jsRedirectPath) {
+                try {
+                  resolvedRequest = await resolver(context, resolvedRequest);
+                } catch (e) {
+                  // Do nothing, fallthrough to other external matches.
+                  logger.debug(
+                    `Failed to resolve ${resolvedRequest} with resolver`,
```
Timeless0911, line 1023: should the error be printed to debug info？

packages/core/src/config.ts
```diff
@@ -955,64 +959,113 @@ const composeEntryConfig = async (
...
-                if (JS_EXTENSIONS_PATTERN.test(request)) {
-                  request = request.replace(/\.[^.]+$/, jsExtension);
+                resolvedRequest = normalizeSlash(
+                  path.relative(
+                    path.dirname(contextInfo.issuer),
+                    resolvedRequest,
+                  ),
+                );
+
+                if (resolvedRequest[0] !== '.') {
```
Timeless0911, line 1034: Why to convert to relative path？

packages/core/src/types/config.ts
```diff
@@ -74,12 +74,33 @@ export type Shims = {
...
+// @ts-expect-error TODO: support dts redirect in the future
+type DtsRedirect = {
+  path?: boolean;
+  extension?: boolean;
+};
+
 export type Redirect = {
-  // TODO: support other redirects
-  // alias?: boolean;
+  /** Controls the redirect of the import paths of output JavaScript files. */
```
Timeless0911, line 97: not only import paths?

website/docs/en/config/lib/redirect.mdx
```diff
@@ -1,9 +1,23 @@
 # lib.redirect
 
+:::info
+
+Redirect is the specific configuration for bundleless mode (set [lib.bundle](/config/lib/bundle) to `false`). It will not take effect in bundle mode because, all output files are packaged into a single file. As a result, import paths do not exist and do not need to be redirected.
```
Timeless0911, line 5:
```suggestion
Redirect is the unique configuration for bundleless mode (set [lib.bundle](/config/lib/bundle) to `false`). It will not take effect in bundle mode where all output files are packaged into a single file, eliminating the need for import path redirection.
```

website/docs/en/config/lib/redirect.mdx
```diff
@@ -12,39 +26,82 @@ type Redirect = {
...
 ```
 
-Configure the redirect of the import paths.
+Configure the redirect for import paths in output files. In bundleless mode, there are often needs such as using aliases or automatically appending suffixes for ESM products. The `redirect` configuration is designed to address these issues.
 
-When `bundle: false`, the import path is redirected to ensure that it points to the correct output.
+Common scenarios that require redirect:
 
-:::note
+- Automatically convert tsconfig.json paths to correct relative path
```
Timeless0911, line 41:
```suggestion
- Automatically convert `paths` in `tsconfig.json` to correct relative path
```

packages/core/src/config.ts
fi3ework: I'm not sure which request failed to resolve. Bundleless is similar to transform, so regarding transform, we might be able to ignore the error for now.

packages/core/src/config.ts
fi3ework: If a request is falling to here, it means it doesn't match any of the external rule. Which means it fall through the first three externals config to the `bundlelessExternalConfig`. Which will be treated as source code.

- autoExternalConfig,
- targetExternalsConfig,
- userExternalsConfig,
- bundlelessExternalConfig,

packages/core/src/types/config.ts
fi3ework: here the paths means the full path, which is used in other places as well.

packages/core/src/config.ts
Timeless0911: I mean why adding `./` before `resolvedRequest` which is not begon with `.`

fi3ework: It is, https://nodejs.org/api/path.html#pathrelativefrom-to does not add the leading `./`.

Timeless0911:
I think the pr title should be changed.

packages/core/src/config.ts
fi3ework: IMO, #251 means: the rewritten path should be "resovled", which is the issue addressed in this PR. The error to omit here is that a module can't be resolved. They're slightly different.

Timeless0911: @SoonIter What's your opinion?

fi3ework: I think we can first come up with a reasonable use case for this error here. If we can't find one, we can assume that there is no use case, and then we can directly throw an error.

fi3ework: Leave it here for now, we could change this in the future if needed.

Timeless0911 (approved):
Update docs and tests about `externals` in bundleless mode.

noshower:
虽然值的导入解决了别名问题，但是类型导入，生成的 .d.ts还是保留了路径别名

PR #552, opened 2024-12-12
fix: shared Rsbuild plugins should run only once

chenjiahan:
## Summary

I found this issue when testing [rsbuild-plugin-publint](https://github.com/rspack-contrib/rsbuild-plugin-publint).

In Rsbuild, shared plugins will run only once. Take `rsbuild-plugin-publint` as an example, it uses `api.onAfterBuild` to trigger the linting, and this should only be triggered once per build, otherwise it will result in duplicate lint output.

```ts
// rsbuild.config.ts
export default {
  // this is a shared plugin
  plugins: [publintPlugin],
  environments: {
    client: {
      plugins: [clientOnlyPlugin],
    },
    server: {
      plugins: [serverOnlyPlugin],
    },
  },
};
```

Currently Rslib does not align with this behavior, the shared plugins are registered multiple times. After this PR, Rslib will be aligned with Rsbuild.

## Checklist

### DIFF
--- a/packages/core/src/cli/inspect.ts
+++ b/packages/core/src/cli/inspect.ts
@@ -10,6 +10,7 @@ export async function inspect(
   const environments = await composeRsbuildEnvironments(config);
   const rsbuildInstance = await createRsbuild({
     rsbuildConfig: {
+      plugins: config.plugins,
       environments: pruneEnvironments(environments, options.lib),
     },
   });
--- a/packages/core/src/cli/mf.ts
+++ b/packages/core/src/cli/mf.ts
@@ -25,8 +25,15 @@ async function initMFRsbuild(
   }
 
   mfRsbuildConfig.config = changeEnvToDev(mfRsbuildConfig.config);
+
   const rsbuildInstance = await createRsbuild({
-    rsbuildConfig: mfRsbuildConfig.config,
+    rsbuildConfig: {
+      ...mfRsbuildConfig.config,
+      plugins: [
+        ...(rslibConfig.plugins || []),
+        ...(mfRsbuildConfig.config.plugins || []),
+      ],
+    },
   });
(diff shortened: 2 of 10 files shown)


packages/core/src/cli/inspect.ts
```diff
@@ -10,6 +10,7 @@ export async function inspect(
   const environments = await composeRsbuildEnvironments(config);
   const rsbuildInstance = await createRsbuild({
     rsbuildConfig: {
+      plugins: config.plugins,
```
Timeless0911, line 13: missing `mf dev` commands in `packages/core/src/cli/mf.ts`?

chenjiahan: Nice catch, this has been fixed

packages/core/src/cli/mf.ts
```diff
@@ -25,8 +25,15 @@ async function initMFRsbuild(
...
 
   mfRsbuildConfig.config = changeEnvToDev(mfRsbuildConfig.config);
+
   const rsbuildInstance = await createRsbuild({
-    rsbuildConfig: mfRsbuildConfig.config,
+    rsbuildConfig: {
+      ...mfRsbuildConfig.config,
+      plugins: [
+        ...(rslibConfig.plugins || []),
+        ...(mfRsbuildConfig.config.plugins || []),
```
Timeless0911, line 34: Why `mfRsbuildConfig.config.plugins` is added to shared plugins?

chenjiahan: This is not the shared plugins, it's the merged final plugins

Timeless0911: Got it, we should refactor this to using `environments` as same as `build` and `inspect` command.

Timeless0911 (approved):

PR #581, opened 2024-12-19
test: add alias replace externals tests in bundleless mode

Timeless0911:
## Summary

In bundleless mode, since there is no bundling concept, all modules will be externalized, so we will automatically replace the aliases in `resolve.alias` with `output.externals` to ensure the final output is generated correctly.

## Checklist


fi3ework (commented):
Do we need to also add new test cases that rewritten with `resolve.alias` of the cases under `tests/integration/externals/browser` which is using `output.externals` now, since we're might encourage user to use `resolve.alias` first in the future.

fi3ework (approved):

Issue #583, opened 2024-12-19
[Bug]: rslib duplicates public folder contents by default; no option found to disable. Multiple libs cause public contents to be copied to sibling folders.


superchenwb:
### Version

```sh
System:
    OS: Windows 11 10.0.26100
    CPU: (32) x64 Intel(R) Core(TM) i9-14900HX
    Memory: 7.51 GB / 31.73 GB
  Browsers:
    Edge: Chromium (130.0.2849.56)
  npmPackages:
    @rslib/core: ^0.1.5 => 0.1.5
```

### Details

rslib will, by default, copy all the contents of the public folder, and there is no found way to disable this copying. Moreover, when there are multiple libs, it will copy all the contents of the public folder to a folder that is at the same level as the public folder.

This seems to be a behavior inherited from rsbuild.

But I don't need rslib to handle these files; they are not used anywhere in the code, yet I need to keep them.

### Reproduce link

https://github.com/superchenwb/rslib-demo

### Reproduce Steps

for example：

execute pnpm dev
find images folder in the root folders

Timeless0911:
> no option found to disable.

Yeah, this is a behavior inherited from Rsbuild. You can set as below to disable this behaviour.

```ts
server: {
  publicDir: {
    copyOnBuild: false,
  },
},
```

But `server` config is not processed in Rslib, I will fix it soon.

> Multiple libs cause public contents to be copied to sibling folders.

This maybe an Rsbuild bug, @9aoy cc

NexxLuo:
rslib更新到0.2.0，当配置了`source.assetsInclude`后，会将相应文件复制到静态资源目录中，并且并没有生成源文件对应的目录而是单纯将文件复制，从而导致报错`Conflict: Multiple assets emit different content to the same filename static/assets/style.scss`。
是否有方法禁用？

Timeless0911:
> rslib更新到0.2.0，当配置了`source.assetsInclude`后，会将相应文件复制到静态资源目录中，并且并没有生成源文件对应的目录而是单纯将文件复制，从而导致报错`Conflict: Multiple assets emit different content to the same filename static/assets/style.scss`。 是否有方法禁用？

Please open another issue and provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to help us address the issue, we cannot assist with troubleshooting based on the information currently provided.
Thanks.

[Why reproduction is required](https://antfu.me/posts/why-reproductions-are-required)

[...]

PR #584, opened 2024-12-19
fix: pass server config to Rsbuild

Timeless0911:
## Summary

We should pass `server` config to Rsbuild since `mf dev` and some `publicDir` feature need to configure it.

## Related Links

fix: #583 

## Checklist

### DIFF
--- a/packages/core/src/cli/build.ts
+++ b/packages/core/src/cli/build.ts
@@ -12,6 +12,7 @@ export async function build(
   const rsbuildInstance = await createRsbuild({
     rsbuildConfig: {
       plugins: config.plugins,
+      server: config.server,
       environments: pruneEnvironments(environments, options.lib),
     },
   });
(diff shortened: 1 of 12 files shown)


9aoy (approved):

fi3ework (commented):
I think we need to document out public dir https://rsbuild.dev/zh/guide/basic/static-assets#public-%E7%9B%AE%E5%BD%95 in Rslib document as well. For now, there's not any word for it, and user won't know it without reading Rsbuild's doc.

Timeless0911:
> I think we need to document out public dir https://rsbuild.dev/zh/guide/basic/static-assets#public-%E7%9B%AE%E5%BD%95 in Rslib document as well. For now, there's not any word for it, and user won't know it without reading Rsbuild's doc.

We can introduce it in assets chapter after we support all assets feature well.

fi3ework (approved):

Issue #555, opened 2024-12-13
Refactor compose Rsbuild config function in mf dev command


Timeless0911:
Got it, we should refactor this to using `environments` as same as `build` and `inspect` command.

_Originally posted by @Timeless0911 in https://github.com/web-infra-dev/rslib/pull/552#discussion_r1883259580_

Issue #513, opened 2024-12-03
[Feature]: Support `--lib` param of `rslib mf dev` command


Timeless0911:
### What problem does this feature solve?

Support `--lib` param of `rslib mf dev` command

### What does the proposed API look like?

The same as `rslib build --lib`.

PR #590, opened 2024-12-22
feat!: support `--lib` param of `mf-dev` command

Timeless0911:
## Breaking change

- Modify `mf dev` command name to `mf-dev` to accect more params

## Summary

- Support `--lib` param of `mf-dev` command
- Use `EnvironmentConfig` type in `lib` field
- Use `EnvironmentConfig` instead of `RsbuildConfig` types in composing Rsbuild config in Rslib internal logic
- Put below config to outside of `environments` config since they are global configuration of Rsbuild
	- `mode`
	- `root`
	- `dev`
	- `plugins`
	- `server`
- Should write `server` config in shared Rsbuild configurations instead of object of `lib` array which represents environments config

## Related Links

closes: #513 
closes: #555 

## Checklist


packages/core/src/cli/mf.ts
```diff
@@ -1,42 +1,63 @@
...
-      server: mergeRsbuildConfig(
-        rslibConfig.server,
-        mfRsbuildConfig.config.server,
-      ),
+      mode: 'development',
+      root: config.root,
+      plugins: config.plugins,
+      dev: {
+        ...(config.dev ?? {}),
+        writeToDisk: true,
```
fi3ework, line 49: Why does MF didn't declare this field before?

Timeless0911: image

fi3ework: Got it, I'm blind.

fi3ework (approved):

PR #607, opened 2024-12-25
fix(create-rslib): remove unnecessary React import

fi3ework:
## Summary

- Use `import type` for `React` when we're already using `automatic` jsx runtime
- Use `exclude` in `tsconfig.json` to make the test files are also included and get correct type resolve.

## Checklist


packages/create-rslib/fragments/base/react-ts/tsconfig.json
```diff
@@ -10,5 +10,5 @@
     "moduleResolution": "bundler",
     "useDefineForClassFields": true
   },
-  "include": ["src"]
+  "exclude": ["node_modules"]
```
Timeless0911, line 13: Will this change affect the dts results? The best practice should be that the outer `tsconfig.json` is responsible for all type verification, and there is a `tsconfig.build.json` responsible for emitting dts files.

fi3ework: Agreed, and, without using project reference?

fi3ework: You're right, dts messed up, a lot of config file emitted. 🥲

Timeless0911: `the test files are also included and get correct type resolve`  I think this should configure in user side. Our tpl can ignore this.

packages/create-rslib/template-[react]-[storybook,vitest]-ts/tests/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "../tsconfig.json",
+  "include": [".", "../vitest.setup.ts"]
```
Comment (Timeless0911) [inline packages/create-rslib/template-[react]-[storybook,vitest]-ts/tests/tsconfig.json:3]: the storybook tpl can also add a separated `tsconfig.json`?

Timeless0911 (approved):

PR #630, opened 2024-12-30
test: enable persistent cache

Timeless0911:
## Checklist


Timeless0911:
image

packages/core/src/config.ts
```diff
@@ -1474,7 +1479,20 @@ export async function composeRsbuildEnvironments(
 
   for (const { format, id, config } of rsbuildConfigWithLibInfo) {
     const libId = typeof id === 'string' ? id : composeDefaultId(format);
-    environments[libId] = config;
+    environments[libId] = mergeRsbuildConfig(config, {
+      tools: {
+        rspack: {
+          experiments: {
+            cache: {
+              version:
```
Timeless0911: invalidate:
- `config.name`
- config hash

Timeless0911:
`LIB_CSS_EXTRACT_LOADER` emit should set `this.cacheable(false)`.

Timeless0911:
Run `pnpm test:integration` twice
1. Vue case
image
2. External module warning case -- external function does not be executed since hitting cache
image
3. watch case
image

Timeless0911:
image
seems not work fine with multi compiler for cache invalidate key

If different config with same id, it will fail too, see this case https://github.com/web-infra-dev/rslib/blob/test/persis/tests/integration/shims/esm/rslibShimsDisabled.config.ts

Timeless0911:
There is only one case left now, which is related to the implementation of unplugin.

image

Issue #632, opened 2024-12-31
[Bug]: Error occurred when packaging UMD files  './src/**?__rslib_entry__  not found


yoke0104x:
### Version

```sh
System:
    OS: macOS 15.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 172.78 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 131.0.6778.205
    Safari: 18.2
  npmPackages:
    @rslib/core: ^0.2.2 => 0.2.2
```

### Details

![Image](https://github.com/user-attachments/assets/543f2364-c2af-4543-a342-2505e2d7d3aa)

Package as a component src/index as the main entry component
If you set the index to ./src/index.tsx, there will be no error (there is no content in ESM, CJS)
If you set the index to ./src/**, there will be a problem with the above screenshots
```
import { pluginReact } from '@rsbuild/plugin-react';
import { defineConfig } from '@rslib/core';

export default defineConfig({
  source: {
    entry: {
      index: ['./src/**'],
    },
  },
  lib: [
    {
      bundle: false,
      dts: true,
      format: 'esm',
      output: {
        distPath: {
          root: './dist/esm',
        },
        minify: true,
      },
    },
    {
      bundle: false,
      dts: true,
      format: 'cjs',
      output: {
        distPath: {
          root: './dist/cjs',
        },
        minify: true,
      },
    },
    {
      format: 'umd',
      umdName: 'rsCom',
      output: {
        externals: {
          react: 'rsCom',
        },
        distPath: {
          root: './dist/umd',
        },
      },
    },
  ],
  output: {
    target: 'web',
  },
  plugins: [pluginReact()],
});

```

### Reproduce link

https://github.com/yoke0104x/temp.git

### Reproduce Steps

simple steps
1. cd rs_com
2. pnpm i
3. pnpm build
4. you can see the error message

Timeless0911:
Glob pattern entry like `.src/**` must be set along with `bundle: false`.

yoke0104x:
![Image](https://github.com/user-attachments/assets/a1770150-8d5a-4738-be18-792ac6f61202)

If set in the umd option `bundle: false`, this error will be reported .

yoke0104x:
> Glob pattern entry like `.src/**` must be set along with `bundle: false`.

![Image](https://github.com/user-attachments/assets/a1770150-8d5a-4738-be18-792ac6f61202)

[...]

PR #634, opened 2024-12-31
fix: detect possible glob pattern in bundle more

fi3ework:
## Summary

Close https://github.com/web-infra-dev/rslib/issues/632.

> > Glob pattern entry like .src/** must be set along with bundle: false.
> 
> I think we could do an active detection here.

image

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -890,8 +890,11 @@ const composeEntryConfig = async (
   if (bundle !== false) {
     let isFileEntry = false;
     traverseEntryQuery(entries, (entry) => {
-      if (fs.existsSync(entry)) {
-        const stats = fs.statSync(entry);
+      const entryAbsPath = path.isAbsolute(entry)
+        ? entry
+        : path.resolve(root, entry);
+      if (fs.existsSync(entryAbsPath)) {
+        const stats = fs.statSync(entryAbsPath);
         isFileEntry = stats.isFile();
       }
 
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -3,6 +3,7 @@ import path, { dirname, extname, isAbsolute, join } from 'node:path';
 import {
   type EnvironmentConfig,
   type RsbuildConfig,
+  type RsbuildEntry,
   type RsbuildPlugin,
   type RsbuildPlugins,
   type Rspack,
@@ -844,8 +845,9 @@ const composeSyntaxConfig = (
   };
 };
 
-export const appendEntryQuery = (
+const traverseEntryQuery = (
   entry: RsbuildConfigEntry,
+  callback: (entry: string) => string,
 ): RsbuildConfigEntry => {
   const newEntry: Record<string, RsbuildConfigEntryItem> = {};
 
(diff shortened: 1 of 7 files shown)


packages/core/src/config.ts
```diff
@@ -883,6 +888,20 @@ const composeEntryConfig = async (
   }
 
   if (bundle !== false) {
+    let hasGlobPattern = false;
+    traverseEntryQuery(entries, (entry) => {
+      if (entry.includes('*')) {
```
Timeless0911, line 893:
```suggestion
      if (/[*?]/.test(entry)) {
```

packages/core/src/config.ts
```diff
@@ -883,6 +888,20 @@ const composeEntryConfig = async (
...
+        hasGlobPattern = true;
+      }
+      return entry;
+    });
+
+    if (hasGlobPattern) {
+      throw new Error(
+        `glob pattern is not supported when "bundle" is "true", considering ${color.green('set "bundle" to "false"')} to use bundleless mode. See ${color.green('https://lib.rsbuild.dev/guide/basic/output-structure')} for more details.`,
```
Timeless0911, line 901:
```suggestion
        `Glob pattern is not supported when "bundle" is "true", considering ${color.green('set "bundle" to "false"')} to use bundleless mode. See ${color.green('https://lib.rsbuild.dev/guide/basic/output-structure')} for more details.`,
```

tests/integration/entry/index.test.ts
```diff
@@ -98,3 +99,17 @@ test('glob entry bundleless', async () => {
...
+  const fixturePath = join(__dirname, 'glob-bundle');
+  let errMsg = '';
+  try {
+    await buildAndGetResults({ fixturePath });
+  } catch (e) {
+    errMsg = (e as Error).message;
+  }
+
+  expect(stripAnsi(errMsg)).toMatchInlineSnapshot(
+    `"glob pattern is not supported when "bundle" is "true", considering set "bundle" to "false" to use bundleless mode. See https://lib.rsbuild.dev/guide/basic/output-structure for more details."`,
```
Timeless0911, line 113:
```suggestion
    `"Glob pattern is not supported when "bundle" is "true", considering set "bundle" to "false" to use bundleless mode. See https://lib.rsbuild.dev/guide/basic/output-structure for more details."`,
```

Timeless0911:
Should we add `./src/**/*` as default entry of bundleless mode?

packages/core/src/config.ts
```diff
@@ -883,6 +888,22 @@ const composeEntryConfig = async (
   }
 
   if (bundle !== false) {
+    let isFileEntry = false;
+    traverseEntryQuery(entries, (entry) => {
+      if (fs.existsSync(entry)) {
```
Timeless0911, line 893: Does absolute/relative paths matter here? And what if users set [root](https://rsbuild.dev/zh/config/root#root)?

fi3ework: Nice catch, changed to abs path to test the existence.

Timeless0911 (approved):

PR #642, opened 2025-01-04
fix: handle add and unlink file in bundleless mode

fi3ework:
## Summary

Previously, the `entry` in bundleless mode is set during the initial build and does not change during the `rslib build --watch` phase. For example, a newly added file will not be monitored, while a file that was included initially will trigger an error (entry module not found) if it is deleted later.

In this PR, we use two combined functionality to resolve the issue:

1. Dynamic entry (https://rspack.dev/config/entry#dynamic-entry): a callback that will revoked when watch build is triggered every time that give us a chance to re-scan the entires.
2. `compilation.contextDependencies`: an extra directory to watch and trigger watch build (e.g. used in https://github.com/soda-x/extra-watch-webpack-plugin)

Limitation:

- `compilation.contextDependencies` will be resolved at the very beginning by LCP path, it won't be changed anymore.

## Checklist

### DIFF
--- a/packages/core/src/plugins/EntryChunkPlugin.ts
+++ b/packages/core/src/plugins/EntryChunkPlugin.ts
@@ -37,40 +37,47 @@ class EntryChunkPlugin {
   private shebangInjectedAssets: Set<string> = new Set();
 
   private enabledImportMetaUrlShim: boolean;
+  private contextToWatch: string | null = null;
+  private contextWatched = false;
 
   constructor({
     enabledImportMetaUrlShim = true,
+    contextToWatch,
   }: {
     enabledImportMetaUrlShim: boolean;
+    contextToWatch: string | null;
   }) {
     this.enabledImportMetaUrlShim = enabledImportMetaUrlShim;
+    this.contextToWatch = contextToWatch;
   }
 
   apply(compiler: Rspack.Compiler) {
-    compiler.hooks.entryOption.tap(PLUGIN_NAME, (_context, entries) => {
-      for (const name in entries) {
-        const entry = (entries as Rspack.EntryStaticNormalized)[name];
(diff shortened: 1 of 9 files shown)


packages/core/src/plugins/EntryChunkPlugin.ts
```diff
@@ -87,7 +93,25 @@ class EntryChunkPlugin {
       }
     });
 
-    compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
+    compiler.hooks.make.tap(PLUGIN_NAME, (compilation) => {
```
Timeless0911, line 96: this logic has been implemented twice?

tests/integration/directive/react/bundleless/rslib.config.ts
```diff
@@ -1,22 +1,22 @@
...
       bundle: false,
       output: {
         distPath: {
-          root: './dist/bundle/esm0',
+          root: './dist/esm',
         },
       },
     }),
-    generateBundleEsmConfig({
+    generateBundleCjsConfig({
```
Timeless0911, line 14: Why this change?

fi3ework: To test CJS output as well, better than test against ESM output twice.

packages/core/src/plugins/EntryChunkPlugin.ts
fi3ework: Changed.

fi3ework: GitHub diff is terrible, you could check it on VS Code GitHub extension.

Timeless0911 (approved):

PR #653, opened 2025-01-07
fix: user externals should have higher priority

fi3ework:
## Summary

Fix bug introduced in https://github.com/web-infra-dev/rslib/pull/535/files#diff-4114b8f08217be27efd35a3891857cfabc6eb75aee8ab39ba304e4d1b1095323R1317.

User's externals config should take higher priority to matched earlier than `autoExternalConfig` and others.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1391,7 +1391,7 @@ async function composeLibRsbuildConfig(
     // #region Externals configs
     // The order of the externals config should come in the following order:
     // 1. `externalsWarnConfig` should come before other externals config to touch the externalized modules first.
-    // 2. `userExternalsConfig` should present later to override the externals config of the ahead ones.
+    // 2. `userExternalsConfig` should present earlier to override the externals config of the ahead ones.
     // 3. The externals config in `bundlelessExternalConfig` should present after other externals config as
     //    it relies on other externals config to bail out the externalized modules first then resolve
     //    the correct path for relative imports.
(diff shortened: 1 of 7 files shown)


packages/core/src/config.ts
```diff
@@ -1391,14 +1391,14 @@ async function composeLibRsbuildConfig(
     // #region Externals configs
     // The order of the externals config should come in the following order:
     // 1. `externalsWarnConfig` should come before other externals config to touch the externalized modules first.
-    // 2. The externals config in `bundlelessExternalConfig` should present after other externals config as
+    // 2. `userExternalsConfig` should present later to override the externals config of the ahead ones.
```
Timeless0911, line 1394: later？

Timeless0911: This is because in bundleless mode, we return callbacks for all requests, so users externals configuration does not takes effect?

fi3ework:
```suggestion
    // 2. `userExternalsConfig` should present earlier to override the externals config of the ahead ones.
```

packages/core/src/config.ts
```diff
@@ -1391,14 +1391,14 @@ async function composeLibRsbuildConfig(
     // #region Externals configs
     // The order of the externals config should come in the following order:
     // 1. `externalsWarnConfig` should come before other externals config to touch the externalized modules first.
-    // 2. The externals config in `bundlelessExternalConfig` should present after other externals config as
+    // 2. `userExternalsConfig` should present earlier to override the externals config of the ahead ones.
```
fi3ework, line 1394:
```suggestion
    // 2. `userExternalsConfig` should present at first to takes effect earlier than others.
```

packages/core/src/config.ts
fi3ework: Wrong desc. The earlier an external present, the higher priority it have.

Timeless0911 (approved):

PR #656, opened 2025-01-07
chore: hoist mf plugin to global

2heal1:
## Summary

hoist mf plugin to global

## Checklist

- [X] Documentation updated (or not required).

### DIFF
--- a/examples/module-federation/mf-react-component/rslib.config.ts
+++ b/examples/module-federation/mf-react-component/rslib.config.ts
@@ -33,22 +33,6 @@ export default defineConfig({
       dev: {
         assetPrefix: 'http://localhost:3001/mf',
       },
-      plugins: [
-        pluginModuleFederation({
-          name: 'rslib_provider',
-          exposes: {
-            '.': './src/index.tsx',
-          },
-          shared: {
-            react: {
-              singleton: true,
-            },
-            'react-dom': {
-              singleton: true,
-            },
-          },
-        }),
-      ],
     },
   ],
   source: {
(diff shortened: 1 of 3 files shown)


fi3ework (commented):
Is this change for functionality reason, or just polish code style?

2heal1:
functionality reason, we may inject variables for runtime api in the future , so the plugin need to be hoisted as global plugin .

2heal1:
> Is this change for functionality reason, or just polish code style?


functionality reason, we may inject variables for runtime api in the future , so the plugin need to be hoisted as global plugin .

fi3ework:
> If not , we can only affect the mf type asset , the other assets will be crashed without injecting our expected variables.

So if there's a lib config which is a pure ESM or CJS output without any MF stuff, how could it bail out from runtime injecting of the plugin?

I guess something like below?

```js
[
  { plugins: [mfPlugin]}, // MF related lib
  { plugins: [mfPlugin]}, // MF related lib
  { plugins: []},                // non-MF related lib
]
```

Timeless0911:
What if there are multiple objects in `lib` array, global plugins will take effect for them all. How to use different options to distinguish them? 

And we have supported `npx rslib mf-dev --lib mf1` to choose one lib of mf format, is this feature not needed?

fi3ework:
@2heal1 Do we need a update?

2heal1:
It only affect format: 'mf' by checking the library.type https://github.com/module-federation/core/blob/main/packages/rsbuild-plugin/src/cli/index.ts#L35

Timeless0911:
What if there are two build configurations with format `mf` in the `lib` array, does the global mf plugin mean they will use the same plugin configuration?

2heal1:
Ahh yes, i forget this scene. The pr should be closed, i need to think this scene, especially i want to add ssr output as well

PR #659, opened 2025-01-07
fix: validate entry correctness in bundle mode

fi3ework:
## Summary

Previously, if the entry is pointing to a non-existed file in bundle mode, it will throw an error that hints `Glob pattern ...` which is wrong. 

In this PR improved the validating of entry, which could determine non-existing file entry and misuse dir entry in bundle mode.

## Checklist


packages/core/src/config.ts
```diff
@@ -899,22 +899,40 @@ const composeEntryConfig = async (
...
+          // Existed file.
+        }
+      } else {
+        if (isDirLike) {
+          // Non-existed dir.
+          entryErrorReasons.push(dirError);
+        } else {
+          // Non-existed file.
+          entryErrorReasons.push(
+            `Can't resolve the entry ${color.red(`"${entry}"`)} at the location ${color.red(`${entryAbsPath}`)}. Please ensure that the file exists.`,
```
Timeless0911, line 925:
```suggestion
            `Can't resolve the entry ${color.cyan(`"${entry}"`)} at the location ${color.cyan(`${entryAbsPath}`)}. Please ensure that the file exists.`,
```

Timeless0911: image
align with Rsbuild

fi3ework: cyan is fine as well.

Timeless0911 (approved):

Issue #674, opened 2025-01-13
[Bug]: Monorepo bundleless mode generates wrongly relative imports


Jungzl:
### Version

```sh
System:
    OS: Windows 11 10.0.26100
    CPU: (24) x64 13th Gen Intel(R) Core(TM) i7-13700
    Memory: 2.24 GB / 15.77 GB
  Browsers:
    Edge: Chromium (131.0.2903.63)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @rslib/core: ^0.3.1 => 0.3.1
```

### Details

In monorepo with `bundle:false`, package-a import package-b will generate relative import path.

### Reproduce link

https://github.com/Jungzl/rslib-project-issue-demo

### Reproduce Steps

1. pnpm build
2. checkout `packages/bar/dist/index.js`

```js
import * as __WEBPACK_EXTERNAL_MODULE__foo_dist_index_js_0021292a__ from "../../foo/dist/index.js";
```

Jungzl:
Can be solved by adding `package-b` to `package-a`'s `output.externals`

Timeless0911:
> Can be solved by adding `package-b` to `package-a`'s `output.externals`

This is a valid workaround solution. But I think this case should be enhanced in Rslib internally. 

This is a good error catching, we will try to fix this.

Timeless0911:
Rslib enable [autoExternal](https://lib.rsbuild.dev/config/lib/auto-external#libautoexternal) by default which will detect dependencies in `package.json`. But it not takes effect in your project since you install `@rslib/core` in the root directory of monorepo instead of sub project.

Timeless0911:
> Can be solved by adding `package-b` to `package-a`'s `output.externals`

It is a valid workaround solution. Let us try to compatible with case that use `@rslib/core` as phamton dependency in monorepo.

Timeless0911:
Oh, sorry for that. I think the actual reason is that you use bundleless mode in which `autoExternal` can not take effect in our design. 

In fact, `foo` is a symlink without `node_modules` path after resolver, we do not handle this situation.

PR #677, opened 2025-01-13
fix: only treat files within context as source file

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/674. Only treat files within project root path as source code and process the redirect.

~~Test cases conflicted, waiting for https://github.com/web-infra-dev/rslib/pull/601 to merged first.~~

## Checklist


packages/core/src/config.ts
```diff
@@ -1083,9 +1090,12 @@ const composeBundlelessExternalConfig = (
                 let resolvedRequest = request;
                 // use resolver to resolve the request
                 resolvedRequest = await resolver!(context!, resolvedRequest);
+                const isSubpath = resolvedRequest.startsWith(
```
Timeless0911, line 1093: Does `resolvedRequest` differ from different platform?

fi3ework: Yes, Windows uses `\\`. https://nodejs.org/api/path.html#pathsep

Timeless0911: seems that rspack resolver keep the origin path?

fi3ework: TBH, I don't know, but it should keep using the sep of the current platform.

Timeless0911 (approved):

PR #683, opened 2025-01-14
feat: clear console before ready log

fi3ework:
## Summary

Previously, Rslib's logs have continuously accumulated results from previous builds, which can make it challenging to observe the outcomes of the most recent builds, particularly when there were errors in earlier builds.

In this PR, soft clear on watch mode.

Before:

https://github.com/user-attachments/assets/2c4dbddf-315f-419f-b7b5-b3f9a69da2b6

After:

https://github.com/user-attachments/assets/48d26830-d71c-4510-85f1-1610dafa5309

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -10,6 +10,7 @@ import {
   defineConfig as defineRsbuildConfig,
   loadConfig as loadRsbuildConfig,
   mergeRsbuildConfig,
+  logger as rsbuildLogger,
   rspack,
 } from '@rsbuild/core';
 import { glob } from 'tinyglobby';
@@ -80,6 +81,15 @@ import {
   transformSyntaxToRspackTarget,
 } from './utils/syntax';
 import { loadTsconfig } from './utils/tsconfig';
+
+rsbuildLogger.override({
+  start: (message) => {
+    // Soft clear, copied from https://github.com/lukeed/console-clear/blob/master/index.js.
+    // "start" callback will only be called on watch mode (first compile not included).
+    process.stdout.write('\x1B[H\x1B[2J');
+    logger.start(message);
+  },
+});
 
 /**
  * This function helps you to autocomplete configuration types.


packages/core/src/config.ts
```diff
@@ -81,6 +82,15 @@ import {
 } from './utils/syntax';
 import { loadTsconfig } from './utils/tsconfig';
 
+rsbuildLogger.override({
+  start: (message) => {
```
chenjiahan, line 86: Overriding start is not a good idea, how about using a hook like `onBeforeBuild`

Timeless0911 (approved):

packages/core/src/config.ts
chenjiahan: We cannot guarantee that logger.start will not be called from somewhere else

fi3ework: I tried `onBeforeBuild` before, but when multiple environments presents. It will be called multiple times. I think I should re-implement something like `Building...` with Rspack `watchRun` and `done` hook in Rslib side.

fi3ework: > onBeforeBuild is a global hook, are you registering plugins in global？

Might be.

chenjiahan: Rsbuild will not clear the logs before rebuilding, as this will likely cause useful logs to be cleared and make troubleshooting more difficult.

fi3ework: I'd like to give a try on Rslib, hoping it fit with the simpler lib scenario. 😀

chenjiahan: In fact, Rsbuild early versions will clear the log by default, but we later found that this caused more trouble than benefit, so we canceled it. However, for Rslib scenarios it is still worth a try.

fi3ework: Can't find an elegant way to archive this, I'm heading to close for now.

- `onBeforeBuild` will run after `Building...` and clear it, so it's not an option.
- `watchRun` is good, but can't find a way to register the plugin for only once. Register to the first environment only is also an ugly workaround.

PR #705, opened 2025-01-24
feat(asset): support svgr exportType usage

SoonIter:
## Summary

[queryImport](https://rsbuild.dev/zh/plugins/list/plugin-svgr#query) feature`?url` or `?react` is not  supported in bundleless mode of Rslib because every files is transformed seperatedly

BTW, both`issuer` and `resourceQuery` are not supported in bundleless mode literally

so svgr has two usages below

**1. `mixedImport: true`**

```ts
 pluginSvgr({
 	mixedImport: true
 })
```

```ts
import url, { ReactComponent } from './assets/logo.svg'
```

**2. `exportType: 'default'` or `exportType: 'named'`**

this is a case when using `exportType: 'default'`

```ts
pluginSvgr({
 	svgrOptions: {
 	 	exportType: 'default',
 	},
 	// excluded svg files will not be transformed by svgr
 	// exclude: /logo2\.svg$/,
}),
```

we currently support the first, this pr add the second usage

## Checklist


Timeless0911 (approved):
Add guides of usagae in assets docs later.

Issue #199, opened 2024-09-12
[Feature]: Support same filename "index.css"  "index.js" in`"bundle": false`


SoonIter:
### What problem does this feature solve?

<img width="500" alt="image" src="https://github.com/user-attachments/assets/7c661aba-28b4-4d3f-a34e-6e57de6347ac">

```ts
import { pluginReact } from '@rsbuild/plugin-react';
import { type LibConfig, defineConfig } from '@rslib/core';

export default defineConfig({
  source: {
    entry: {
      index: ['./src/**', '!./src/env.d.ts'],
    },
  },
  lib: [
    {
      format: 'esm',
      bundle: false,
    },
  ],
  plugins: [pluginReact()],
});

```

<img width="500" alt="image" src="https://github.com/user-attachments/assets/79254b33-f787-4f40-81d6-82ad5f1b20a8">

### What does the proposed API look like?

fix this

fi3ework:
@SoonIter Just test it locally, I think this is resolved now.

SoonIter:
> @SoonIter Just test it locally, I think this is resolved now.

I handle the css file， but .svg and svgr files would meet this case too

PR #715, opened 2025-01-27
fix: warn duplicate entries in bundleless mode

Timeless0911:
## Summary

For duplicate entries in bundleless mode, we already deal with css files which is common to have same name in library development. But for some static assets, which usually not appear at the same level of Javascript files, this PR log a warning if they are in the same level as what Javascript files locates.

## Related Links

close: #199 

## Checklist

### DIFF
--- a/tests/integration/entry/duplicate/rslib.config.ts
+++ b/tests/integration/entry/duplicate/rslib.config.ts
@@ -1,0 +1,10 @@
+import { defineConfig } from '@rslib/core';
+import { generateBundleEsmConfig } from 'test-helper';
+
+export default defineConfig({
+  lib: [
+    generateBundleEsmConfig({
+      bundle: false,
+    }),
+  ],
+});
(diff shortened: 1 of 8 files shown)


tests/integration/entry/duplicate/rslib.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from '@rslib/core';
+import { generateBundleEsmConfig } from 'test-helper';
+
+export default defineConfig({
+  lib: [
+    generateBundleEsmConfig({
+      bundle: false,
```
fi3ework, line 7: By default, assets will be emitted to `./assets` folder. How could this case make asset to conflict with others?

Timeless0911: The js file of assets which is a intermediary to make outputs look as same as source code.

Timeless0911: image

fi3ework (approved):

PR #730, opened 2025-02-04
feat(core): add configuration validation using typia

colinaaa:
> [!IMPORTANT]
> Note that this is currently just a POC implementation.

## Summary

Add a validation to `loadConfig` to make sure that the configuration is correct when loaded.

I'm using [`typia`](https://github.com/samchon/typia) and [`typia-rspack-plugin`](https://github.com/colinaaa/typia-rspack-plugin) to make validation.

1. It use pure TypeScript definition as the source of truth
2. It is fast

## Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -56,7 +56,9 @@
     "rslib": "npm:@rslib/core@0.4.0",
     "rslog": "^1.2.3",
     "tsconfck": "3.1.4",
-    "typescript": "^5.7.3"
+    "typescript": "^5.7.3",
+    "typia": "^7.6.2",
+    "typia-rspack-plugin": "^1.0.1"
   },
   "peerDependencies": {
     "@microsoft/api-extractor": "^7",
--- a/packages/core/rslib.config.ts
+++ b/packages/core/rslib.config.ts
@@ -3,6 +3,7 @@ import path from 'node:path';
 import type { RsbuildPlugin } from '@rsbuild/core';
 import { pluginPublint } from 'rsbuild-plugin-publint';
 import { defineConfig } from 'rslib';
+import { TypiaRspackPlugin } from 'typia-rspack-plugin';
 
 const pluginFixDtsTypes: RsbuildPlugin = {
   name: 'fix-dts-types',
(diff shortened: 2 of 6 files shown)


packages/core/package.json
```diff
@@ -56,7 +56,9 @@
     "rslib": "npm:@rslib/core@0.4.0",
     "rslog": "^1.2.3",
     "tsconfck": "3.1.4",
-    "typescript": "^5.7.3"
+    "typescript": "^5.7.3",
+    "typia": "^7.6.2",
```
chenjiahan, line 60: `typia` has so many dependencies, this is a bit hard to accept...
image

https://pkg-graph.info/typia

colinaaa: It indeed has too many dependencies, but it is tree-shakable and we are using it as `devDependencies`.

The bundle size of Rslib is increased from **106.8 kB** to **2484.4 kB**. But the size change mainly produced by the validation functions generated by `typia`:

image

As you can see, the `src/validate.ts` has **2.33 MB** in stats. And we can make validation opt-in and lazy loaded.

For end users, no additional dependencies need to be installed, and the bundle size of `@rslib/core` increases by only approximately 2.5 MB.

chenjiahan: If we make the validation opt-in, how can users enable the validation?

colinaaa: E.g.: if the `TypiaRspackPlugin`(which transforms `src/validation.ts`) is commented out, the bundle size of `@rslib/core` in this PR is only **108.4 kB**.

image

colinaaa: > If we make the validation opt-in, how can users enable the validation?

Maybe through a environment variable?

```js
// rslib.config.js
process.env['RSLIB_CONFIG_VALIDATION'] = true

export default {}
```

Or maybe a `strictDefineConfig` API?

```js
import {strictDefineConfig} from '@rslib/core'

export default strictDefineConfig({})
```

But I would prefer enabling by default. And opt-out using environment variables.

chenjiahan: If it's opt-in, I don't think most users will actively enable it...

If it's opt-out, the performance overhead doesn't seem worth it, we can write code to do some validation manually.

colinaaa: Sure, I'm closing this PR. Thank you for your review

Issue #714, opened 2025-01-27
[Bug]: No commonjs export names annotations generated


hzy:
## Version

```sh
System:
    OS: Linux 5.15 veLinux GNU/Linux 2 (lyra) 2 (lyra)
    CPU: (56) x64 Intel(R) Xeon(R) Platinum 8336C CPU @ 2.30GHz
    Memory: 99.78 GB / 109.89 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  npmPackages:
    @rslib/core: 0.3.2 => 0.3.2
```

### Background
My package need to be bundled and must has commonjs format, I don't want to double the publish size so "Ship ESM & CJS in one Package
" was not not taken into consideration.

### Problem
When bundle with **esbuild**, a annotation will be generated:

```js
// Annotate the CommonJS export names for ESM import in node:
0 && (module.exports = {
  reactLynxPlugin
});

```

I did not find this kind annotation in rslib output.

This will cause problems:

1. can not use named import from esm
    ```text
    import { reactLynxPlugin } from './speedy-plugin/dist/index.js';
             ^^^^^^^^^^^^^^^
    SyntaxError: Named export 'reactLynxPlugin' not found. The requested module './speedy-plugin/dist/index.js' is a CommonJS module, which may not support all module.exports as named exports.
    CommonJS modules can always be imported via the default export, for example using:
    
    import pkg from './speedy-plugin/dist/index.js';
    const { reactLynxPlugin } = pkg;
    
        at ModuleJob._instantiate (node:internal/modules/esm/module_job:180:21)
        at async ModuleJob.run (node:internal/modules/esm/module_job:263:5)
        at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:547:26)
        at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)
    
    Node.js v22.12.0
    ```
2. some lint tool like attw will fail
   ```text
    🕵️ TypeScript allows ESM named imports of the properties of this CommonJS module, but they will crash at runtime because they don’t exist or can’t be statically detected by Node.js in the JavaScript file. Use -f json to see the list of exports TypeScript can see but Node.js cannot. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NamedExports.md
    
    
    ┌───────────────────┬───────────────────────────────────┐
    │                   │ "@xxxx-xxxx/xxxxx-plugin-react" │
[...]

Issue #351, opened 2024-10-31
[Bug]: The CommonJS bundle can not be consumed by `import()` in Node.js


chenjiahan:
### Version

  System:
    OS: macOS 14.6.1
    CPU: (16) arm64 Apple M3 Max
    Memory: 8.00 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 130.0.6723.71
    Safari: 17.6
  npmPackages:
    @rslib/core: ^0.0.15 => 0.0.15

### Details

- Input:

```ts
export function example(): string {
  return 'hello world';
}
```

- Config:

```ts
import { defineConfig } from '@rslib/core';

export default defineConfig({
  lib: [
    {
      format: 'cjs',
      syntax: 'es2021',
    },
  ],
  output: { target: 'node' },
});
```

- Create a `test.mjs` file to import() the dist bundle:

```ts
async function main() {
    const { example } = await import('./dist/index.cjs');
    console.log(example);
    // tsup: got "[Function: example]"
    // Rslib: got "undefined"
}
main();
```

### Reproduce link

https://github.com/chenjiahan/rslib-repro-cjs-export

### Reproduce Steps

Rslib:

- pnpm run build
- node ./test.mjs

tsup:

- npx tsup
- node ./test.mjs

chenjiahan:
This is required for ESM import in node:

```js
// Annotate the CommonJS export names for ESM import in node:
0 && (module.exports = {
  example
});
```

fi3ework:
I think we're aligned with esbuild's behavior? 

https://esbuild.github.io/try/#YgAwLjI0LjAALS1idW5kbGUgLS1mb3JtYXQ9Y2pzAGUAZW50cnkudHMAZXhwb3J0IGZ1bmN0aW9uIGV4YW1wbGUoKTogc3RyaW5nIHsKICByZXR1cm4gJ2hlbGxvIHdvcmxkJzsKfQo

```ts
async function main() {
  const thing = await import('./dist/index.cjs')
  console.log(thing, thing.example)
}
```

Checked by manually copy to the dist.

esbuild:

![Image](https://github.com/user-attachments/assets/1fb0eb8e-9544-44e2-bd71-cd3e52d3ae1d)


rslib:

![Image](https://github.com/user-attachments/assets/bca5c9e2-7465-402a-89e5-75d9f5fc74ad)

chenjiahan:
[...]

PR #819, opened 2025-03-05
feat!: change to `commonjs-static` library type of commonjs format

fi3ework:
## Summary

`commonjs-static` will directly skip all unprovided exports, which will lead to lose exports. We need to fix this in webpack / Rspack: add `commonjs` like runtime code to export dynamically as the fallback of unprovided exports.

## Related Links

fix: #351 
fix: #714 

## Checklist


Timeless0911 (approved):

Timeless0911:
So much snapshot test cases related to cjs runtime tpl.

PR #820, opened 2025-03-06
fix: should check lib array non-empty

Timeless0911:
## Summary

`lib` field is **required** in Rslib, we should check it is a valid array with at least one element.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1571,9 +1571,11 @@ export async function composeCreateRsbuildConfig(
     ...sharedRsbuildConfig
   } = rslibConfig;
 
-  if (!libConfigsArray) {
+  if (!Array.isArray(libConfigsArray) || libConfigsArray.length === 0) {
     throw new Error(
-      `Expect lib field to be an array, but got ${libConfigsArray}.`,
+      `Expect "lib" field to be a non-empty array, but got: ${color.cyan(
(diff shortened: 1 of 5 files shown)


packages/core/src/config.ts
```diff
@@ -1571,9 +1571,11 @@ export async function composeCreateRsbuildConfig(
     ...sharedRsbuildConfig
   } = rslibConfig;
 
-  if (!libConfigsArray) {
+  if (!Array.isArray(libConfigsArray) || libConfigsArray.length === 0) {
```
fi3ework, line 1574:
```suggestion
  if (Array.isArray(libConfigsArray) && libConfigsArray.length > 0) {
```

I think this way more easy to understand.

Timeless0911: `!Array.isArray(libConfigsArray) || libConfigsArray.length === 0` is equal to `!(Array.isArray(libConfigsArray) && libConfigsArray.length > 0)`.

Timeless0911: 不是数组或数组里没有元素 -> 报错

fi3ework: Yes, identical. Just avoid using `!` if possible.

fi3ework: Nevermind, quite the same.

fi3ework (approved):

PR #837, opened 2025-03-12
test: refactor some style tests

Timeless0911:
## Summary

- remove `root` to avoid warning logs of failed to load `package.json`
- no need to set `lightningcssLoader: false` when using postcss
- resolve `Sass variables aren't allowed in plain CSS` issue in tests
- use `postcss.config.cjs` instead of `postcss.config.ts` to avoid warning in higher Node.js version
- await the assertion to remove unexpected feature which will be removed in Vitest 3.0

## Checklist


SoonIter (approved):
thanks❤

PR #857, opened 2025-03-18
fix: override default values of some type descriptions

quanquan2100:
## Summary

Fix inaccurate type descriptions in `RslibConfig` using `@override` annotation

## Related Links

GH-856

## Checklist

### DIFF
--- a/packages/core/src/types/config.ts
+++ b/packages/core/src/types/config.ts
@@ -304,6 +304,10 @@ export interface LibConfig extends EnvironmentConfig {
    * @see {@link https://lib.rsbuild.dev/config/lib/out-base}
    */
   outBase?: string;
+  /**
+   * @inheritdoc
+   */
+  output?: RslibOutputConfig;
 }
 
 export type LibOnlyConfig = Omit<LibConfig, keyof EnvironmentConfig>;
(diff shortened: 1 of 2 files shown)


packages/core/src/types/config.ts
```diff
@@ -303,8 +308,33 @@ export interface LibConfig extends EnvironmentConfig {
...
+   * @override
+   * @default false
+   */
+  filenameHash?: OutputConfig['filenameHash'];
+  /**
+   * @override
+   * When minify is not specified, Rslib will use a sane default for minify options.
+   * The default options will only perform dead code elimination and unused code elimination.
+   *
+   * detail: {@link https://lib.rsbuild.dev/config/rsbuild/output#outputminify rsbuild-outputminify}
```
Timeless0911, line 327:
```suggestion
   * @see {@link https://lib.rsbuild.dev/config/rsbuild/output#outputminify rsbuild-outputminify}
```

Timeless0911:
`export interface LibConfig extends EnvironmentConfig {`  This line use the type `EnvironmentConfig`, we also need to override in this type when we configure `output.target` in `lib` array.

Timeless0911 (approved):
LGTM, thank you!

Issue #821, opened 2025-03-06
[Feature]: Support yaml and toml


SoonIter:
### What problem does this feature solve?

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/98ad49db-b9c0-452c-abc1-3ad98cf54caf" />

we currently have already supported json. Yaml, toml can be supported in same way but it is not documented right now

### What does the proposed API look like?

https://rsbuild.dev/guide/basic/json-files#yaml-file

https://lib.rsbuild.dev/zh/guide/advanced/json-files

```ts
// rslib.config.ts
import { pluginYaml } from '@rsbuild/plugin-yaml';

plugins: [pluginYaml()]

```

1. support yaml by `pluginYaml`
2. support toml by `pluginToml`

:) Looking forward to someone kind to bring it to fruition. Maybe in two pr with test cases and documents

PR #876, opened 2025-03-26
test: add YAML and TOML tests

Timeless0911:
## Summary

add YAML and TOML tests

## Related Links

close: #821 

## Checklist


tests/integration/asset/index.test.ts
```diff
@@ -258,18 +258,36 @@ test('set the assets public path', async () => {
...
+test('use json/yaml/toml', async () => {
   const fixturePath = join(__dirname, 'json');
   const { contents } = await buildAndGetResults({ fixturePath });
 
   // 0. bundle
   // esm
-  const { content: indexJs } = queryContent(contents.esm0!, /index\.js/);
-  expect(indexJs).matchSnapshot();
+  const { content: bundleIndexJs } = queryContent(contents.esm0!, /index\.js/);
+  expect(bundleIndexJs).matchSnapshot();
```
fi3ework, line 268: Since we might create a new eco CI workflow for Rspack ecosystem CI, I think it's better to add more behavior test (in this case, it's to expose the value from toml/yaml and expect in test file) along with snapshot test.

In eco-ci workflow, snapshot test will be skipped and the behavior test will help us know whether the test case is broken.

SoonIter (approved):

tests/integration/asset/index.test.ts
Timeless0911: I have updated the test.

fi3ework (approved):

PR #878, opened 2025-03-26
docs: add default module definition systems

Timeless0911:
## Summary

add default module definition systems

## Related Links

https://github.com/web-infra-dev/rslib/issues/714#issuecomment-2686798641

## Checklist


website/docs/zh/config/lib/format.mdx
```diff
@@ -11,3 +11,32 @@
...
+            library: {
+              type: 'commonjs',
+            },
+          },
+        },
+      },
+    },
+  ],
+};
+```
```
fi3ework, line 42: Do we need to tell user how to change the library type?

- "modern-module": no substitute.
- "commonjs-static": can be replaced by `"commonjs"`, but `"commonjs-static"` is surely better than it.
- "umd": very slight difference for `"amd"`, rarely used I think.

I think if we add it here, we also need to tell user

1. when to change the type for what purpose.
2. do not change the type in most cases.

Timeless0911: remove it, too complex to describe the differences between the different library types.

fi3ework (approved):

PR #888, opened 2025-03-31
test: use local vendor instead of unpkg for react umd outputs

Timeless0911:
## Summary

Add umd output of `react` and `react-dom` to local to avoid using `unpkg` thus avoiding issues with the CDN service that cause unstable test results.

image

## Checklist

### DIFF
--- a/tests/e2e/react-component/package.json
+++ b/tests/e2e/react-component/package.json
@@ -3,11 +3,10 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
+    "build:umd": "cd ../../../examples/react-component-umd && pnpm run build",
     "dev:bundle": "../../node_modules/.bin/rsbuild dev --environment=bundle",
     "dev:bundle-false": "../../node_modules/.bin/rsbuild dev --environment=bundleFalse",
-    "dev:umd": "pnpm umd:build && pnpm umd:copy && ../../node_modules/.bin/rsbuild dev --environment=umd",
-    "umd:build": "cd ../../../examples/react-component-umd && pnpm run build",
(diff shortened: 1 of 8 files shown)


tests/e2e/react-component/public/umd/react-dom.development.js
chenjiahan, line 1: Can we depend on react 18 package and copy these files from node_modules?

Timeless0911: Good idea. I will do this.

tests/e2e/react-component/package.json
```diff
@@ -7,11 +7,13 @@
     "dev:bundle-false": "../../node_modules/.bin/rsbuild dev --environment=bundleFalse",
     "dev:umd": "pnpm umd:build && pnpm umd:copy && ../../node_modules/.bin/rsbuild dev --environment=umd",
     "umd:build": "cd ../../../examples/react-component-umd && pnpm run build",
-    "umd:copy": "rm -rf ./public && mkdir -p ./public/umd && cp ../../../examples/react-component-umd/dist/umd/index.js ./public/umd/index.js"
+    "umd:copy": "rm -rf ./public && mkdir -p ./public/umd && cp ../../../examples/react-component-umd/dist/umd/index.js ./public/umd/index.js && cp node_modules/react-18/umd/react.development.js ./public/umd/react.development.js && cp node_modules/react-dom-18/umd/react-dom.development.js ./public/umd/react-dom.development.js"
```
chenjiahan, line 10: Can we use the `output.copy` option of Rsbuild

chenjiahan (approved):

Issue #316, opened 2024-10-22
[Bug]: Multiple entry files packaged esm file import error undefind


tym3515:
### Version

"@rslib/core": "^0.0.12",

{
      format: 'esm',
      output: {
        injectStyles: true,
        asyncChunks:false,
        distPath: {
          root: './dist/esm',
          css: '.',
          cssAsync: '.',
        },
      },

### Details

entry: {
  install: './src/install.js',
  Active: './src/lib/Active.js',
}

### Reproduce link

aAAA

### Reproduce Steps

"devDependencies": {
    "@eslint/js": "^9.12.0",
    "@rsbuild/core": "^1.0.13",
    "@rsbuild/plugin-basic-ssl": "^1.1.1",
    "@rsbuild/plugin-sass": "^1.0.3",
    "@rsbuild/plugin-vue2": "^1.0.1",
    "@rslib/core": "^0.0.12",
    "eslint": "^9.12.0",
    "globals": "^15.11.0",
    "prettier": "^3.3.3"
  }

Timeless0911:
Can you describe clearly and provide a reproduce demo?

fi3ework:
A minimal reproduction with webpack. https://stackblitz.com/edit/node-oapeg8?file=webpack.config.js,package.json,bar.js,dist-single-entry%2Fmain.mjs,README.md

Currently, the main issue is that when there are multiple entries, the bailout reasons between different entries are not isolated. This affects the concatenation results of each entry.

With multiple entries, the `./bar.js` bails out from `main` concatenation module. There is no problem when only a `main` entry.

```diff
asset main.mjs 2.98 KiB [emitted] [javascript module] (name: main)
asset bar.mjs 1.65 KiB [compared for emit] [javascript module] (name: bar)
Entrypoint main 2.98 KiB = main.mjs
Entrypoint bar 1.65 KiB = bar.mjs
runtime modules 1.31 KiB 6 modules
orphan modules 61 bytes [orphan] 1 module
cacheable modules 209 bytes
  ./main.js + 1 modules 183 bytes [built] [code generated]
+    ModuleConcatenation bailout: Cannot concat with ./bar.js: Module is an entry point
  ./bar.js 26 bytes [built] [code generated]
webpack 5.95.0 compiled successfully in 124 ms
```

Timeless0911:
https://github.com/web-infra-dev/rslib/pull/1273

Advanced ESM will output like
```js
// 451.js
[...]

PR #896, opened 2025-04-02
refactor: supersede entry query by entry dedicated loader

fi3ework:
## Summary

Remove query entry, it's hack and may raise other unexpected issues. Supersede it with an entry dedicated loader, which will only executed for entry module.

## Checklist


packages/core/src/config.ts
```diff
@@ -1126,7 +1122,8 @@ const composeEntryConfig = async (
       rspack: {
         entry: async () => {
           const { resolvedEntries } = await scanGlobEntries(false);
-          return appendEntryQuery(resolvedEntries);
+          // return appendEntryQuery(resolvedEntries);
```
Timeless0911, line 1125: can be removed

fi3ework:
```suggestion
```

fi3ework:
This PR turns out doesn't resolve multiple entry modules issue which behaves the same as entry query solution when entry module is circular dependented. But this's way better than entry query solution.

Timeless0911 (approved):

fi3ework:
https://github.com/web-infra-dev/rslib/issues/316 is still not resolved in fact (just workaround for now).

Issue #911, opened 2025-04-10
[Bug]: output.filename does not allow setting filename dynamically using a function


hackwater:
### Version

```sh
System:
  OS: Windows 11 10.0.26100
  CPU: (32) x64 AMD Ryzen 9 7950X3D 16-Core Processor          
  Memory: 14.20 GB / 31.18 GB
Browsers:
  Edge: Chromium (127.0.2651.74)
  Internet Explorer: 11.0.26100.1882
npmPackages:
  @rslib/core: ^0.6.2 => 0.6.2
```

### Details

Trying to set the filename dynamically results in an error when running any rslib function. I suspect there's an issue in [composeCreateRsbuildConfig](https://github.com/web-infra-dev/rslib/blob/main/packages/core/src/config.ts#L1585) when calling `mergeRsbuildConfig` to generate the userConfig, but then [calling it again with the userConfig as the third argument](https://github.com/web-infra-dev/rslib/blob/main/packages/core/src/config.ts#L1621). This results in the merger combining two identical functions into a single array [per the logic of the Rsbuild merge function](https://github.com/web-infra-dev/rsbuild/blob/main/packages/core/src/mergeConfig.ts#L61-L63). This array is eventually passed to path.join, which expects a string (I assume the result of the filename function) and therefore fails.

I'm not sure what the right solution is, which is why I'm not trying to provide a PR; it's possible this is an issue stemming from the mergeRsbuildConfig function, but I can definitely successfully generate my config with a dynamic filename in Rsbuild.

Here's a minimal rslib.config.ts that fails when using the dynamic filename example from the Rsbuild docs:

```typescript
import { defineConfig } from '@rslib/core';

export default defineConfig({
  lib: [
    {
      format: 'esm',
    },
  ],
  output: {
    filename: {
      image: (pathData) => {
        if (pathData.filename?.includes('foo')) {
          return '/foo/[name][ext]';
        }
        return '/bar/[name][ext]';
      },
    },
  },
});
```

The error:
```sh
> rslib inspect --verbose

  Rslib v0.6.2

error   Failed to inspect config.
error   TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
    at Object.join (node:path:1268:7)
    at getMergedFilename (file:///C:/Users/xboxl/WebstormProjects/minimal-repro/node_modules/@rsbuild/core/dist/index.js:2358:86)
[...]

PR #912, opened 2025-04-10
fix: set filename dynamically using a function

Timeless0911:
## Summary

In `composeAutoExtensionConfig` fuction, we should only return config about `output.filename.js` instead of whole `output.filename` object to avoid being merged into an array which Rsbuild will throw an error, see https://github.com/web-infra-dev/rslib/issues/911#issuecomment-2791471763.

BTW, just record this scene in this PR that Rslib support setting function for `output.filename.js` and it will take effect in actual output. But we can not get correct js extension when we redirect js path in bundleless mode and will fallback to the default js extension (related to `autoExtension` config) which may not be consistent with the actual output.

Eg.

```ts
// config
output: {
  filename: {
    js: () => {
      return '[name].mjs';
    },
  },
},

// source
import { foo } from './foo';

// output
import * as __WEBPACK_EXTERNAL_MODULE__foo_js_fdf5aa2d__ from "./foo.js";
```

The actual output is `foo.mjs` but we can not get it and redirect to `foo.js` as a fallback.

So if users want to set function for `output.filename.js`, the `[ext]` should better be equal to the default js extension (related to `autoExtension` config).

## Related Links

fix: #911

## Checklist


packages/core/src/config.ts
```diff
@@ -855,25 +855,29 @@ const composeAutoExtensionConfig = (
...
-    typeof updatedConfig.output?.filename?.js === 'string' &&
-    updatedConfig.output?.filename?.js
-      ? extname(updatedConfig.output.filename.js)
+  const defaultJsFilename = `[name]${hash}${jsExtension}`;
+  const userJsFilename = config.output?.filename?.js;
+
+  // will be returned to use in redirect feature
+  // only support string type for now since we can not get the return value of function
+  const finalJsExtension =
+    typeof userJsFilename === 'string' && userJsFilename
```
fi3ework, line 864: what if userJsFilename set to `_foo_[name][ext]`

Timeless0911: `filename.js` and `filename.css` do not support `[ext]`.

Timeless0911: see https://rsbuild.dev/config/output/filename#template-strings

fi3ework (approved):

PR #938, opened 2025-04-19
fix: should process correct format of dts files

Timeless0911:
## Summary

We should only process dts files of current format, this usally happen on multiple dts builds together in the same dist folder.

This should be a regression after #933 

## Checklist

### DIFF
--- a/packages/plugin-dts/src/utils.ts
+++ b/packages/plugin-dts/src/utils.ts
@@ -378,7 +378,7 @@ export async function processDtsFiles(
     );
   }
 
-  const dtsFiles = await globDtsFiles(dir);
+  const dtsFiles = await globDtsFiles(dir, [`/**/*${dtsExtension}`]);
 
   await Promise.all(
     dtsFiles.map(async (file) => {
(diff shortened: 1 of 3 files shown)


packages/plugin-dts/src/utils.ts
```diff
@@ -486,7 +488,8 @@ export const globDtsFiles = async (dir: string): Promise<string[]> => {
 };
 
 export async function cleanDtsFiles(dir: string): Promise<void> {
-  const allFiles = await globDtsFiles(dir);
+  const patterns = ['/**/*.d.ts', '/**/*.d.cts', '/**/*.d.mts'];
+  const allFiles = await globDtsFiles(dir, patterns);
```
fi3ework, line 492: Why clear all possible dts files instead of using given dts format?

Timeless0911: Users can change the dts format by configuration, we should clean all and generate the correct one I think.

fi3ework (approved):

PR #960, opened 2025-04-29
feat(template): add Vue

fi3ework:
## Summary

Quite like the React template.

## Checklist

### DIFF
--- a/packages/create-rslib/fragments/base/vue-js/rslib.config.ts
+++ b/packages/create-rslib/fragments/base/vue-js/rslib.config.ts
@@ -1,0 +1,15 @@
+import { defineConfig } from '@rslib/core';
+import { pluginUnpluginVue } from 'rsbuild-plugin-unplugin-vue';
+
+export default defineConfig({
+  lib: [
+    {
+      bundle: false,
+      format: 'esm',
+    },
+  ],
+  output: {
+    target: 'web',
+  },
+  plugins: [pluginUnpluginVue()],
+});
--- a/packages/create-rslib/rslib.config.ts
+++ b/packages/create-rslib/rslib.config.ts
@@ -1,7 +1,18 @@
 import { pluginPublint } from 'rsbuild-plugin-publint';
 import { defineConfig } from 'rslib';
+const { execSync } = require('node:child_process');
 
 export default defineConfig({
   lib: [{ format: 'esm' }],
-  plugins: [pluginPublint()],
+  plugins: [
+    pluginPublint(),
+    {
+      name: 'rslib:run-generate-template-hook',
+      setup: (api) => {
+        api.onAfterBuild(() => {
+          execSync('pnpm run generate-templates', { stdio: 'inherit' });
+        });
+      },
+    },
+  ],
 });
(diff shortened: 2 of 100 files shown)


packages/create-rslib/fragments/base/vue-js/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "lib": ["DOM", "ES2021"],
+    "module": "ESNext",
+    "jsx": "react-jsx",
```
Jungzl, line 5: Should be
```ts
    "jsx": "preserve",
    "jsxImportSource": "vue",
```
The same goes for the other tsconfigs

chenjiahan: We can copy from https://github.com/web-infra-dev/rsbuild/blob/main/packages/create-rsbuild/template-vue3-ts/tsconfig.json

fi3ework:
This PR is supposed to be updated.

packages/create-rslib/fragments/base/vue-js/rslib.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from '@rslib/core';
+import { pluginUnpluginVue } from 'rsbuild-plugin-unplugin-vue';
+
+export default defineConfig({
+  lib: [
+    {
+      bundle: false,
+      format: 'esm',
+      output: {
+        target: 'web',
```
Timeless0911, line 10: set target on top level outside the `lib`？

packages/create-rslib/rslib.config.ts
```diff
@@ -1,7 +1,18 @@
 import { pluginPublint } from 'rsbuild-plugin-publint';
 import { defineConfig } from 'rslib';
+const { execSync } = require('node:child_process');
```
Timeless0911, line 3:
```suggestion
```

packages/create-rslib/rslib.config.ts
```diff
@@ -1,7 +1,18 @@
 import { pluginPublint } from 'rsbuild-plugin-publint';
```
Timeless0911, line 1:
```suggestion
import { execSync } from 'node:child_process';
import { pluginPublint } from 'rsbuild-plugin-publint';
```

packages/create-rslib/rslib.config.ts
fi3ework: `execSync` is used in line 13.

packages/create-rslib/fragments/base/vue-js/rslib.config.ts
fi3ework: Since here's only one `lib`, I think they're identical here.

Timeless0911: But if users want to add another lib, they should write this again. We can do better default global configuration as possible.

packages/create-rslib/rslib.config.ts
Timeless0911: I just use `import` instead and reorg the import order

Timeless0911 (approved):

PR #976, opened 2025-05-09
feat: format default to `'esm'`

Timeless0911:
## Summary

In this PR, `lib.format` is default from `undefined` to `'esm'` to simplify user configuration and encourage the use of pure esm package.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1512,7 +1512,7 @@ async function composeLibRsbuildConfig(
   const syntaxConfig = composeSyntaxConfig(target, config?.syntax);
   const autoExternalConfig = composeAutoExternalConfig({
     bundle,
-    format: format,
+    format,
     autoExternal,
     pkgJson,
     userExternals: config.output?.externals,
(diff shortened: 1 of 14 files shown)


packages/core/src/config.ts
```diff
@@ -1466,11 +1468,11 @@ async function composeLibRsbuildConfig(
     umdName,
   } = config;
   const { rsbuildConfig: shimsConfig, enabledShims } = composeShimsConfig(
-    format!,
+    format,
     shims,
   );
   const formatConfig = composeFormatConfig({
-    format: format!,
+    format: format,
```
fi3ework, line 1475:
```suggestion
    format,
```

packages/core/src/config.ts
```diff
@@ -1480,14 +1482,14 @@ async function composeLibRsbuildConfig(
...
-    format!,
+    format,
     config.output?.externals,
   );
   const {
     config: autoExtensionConfig,
     jsExtension,
     dtsExtension,
-  } = composeAutoExtensionConfig(config, autoExtension, pkgJson);
+  } = composeAutoExtensionConfig(config, format, autoExtension, pkgJson);
```
fi3ework, line 1492: `config` variable should be renamed to `rawLibConfig` so the other `format` variables are implied to be **resolved**, we could change it in the future.

fi3ework (approved):

PR #982, opened 2025-05-13
feat(create-rslib): syntax default to esnext

Timeless0911:
## Summary

Set `syntax` to `esnext` in `create-rslib` to align with the default value we set.

## Checklist

### DIFF
--- a/packages/create-rslib/fragments/base/node-dual-js/rslib.config.mjs
+++ b/packages/create-rslib/fragments/base/node-dual-js/rslib.config.mjs
@@ -4,11 +4,11 @@ export default defineConfig({
   lib: [
     {
       format: 'esm',
-      syntax: 'es2021',
+      syntax: 'esnext',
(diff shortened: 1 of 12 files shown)


fi3ework (approved):
FWIW: The `syntax` is set to `esnext` by default. However, we explicitly specify it in the template to make it clearer for users, as `syntax` is an important field for them to notice.

chenjiahan:
So why defaults to `esnext`?

Timeless0911:
> So why defaults to `esnext`?

We are a modern build tool I think. What do you think of the default value of `syntax`?

chenjiahan:
For Node.js outputs, I prefer to align with the [Node.js Maintenance LTS](https://nodejs.org/en/about/previous-releases), currently Node.js 18.

When Node 18 reaches EOL, we can raise the default syntax to match Node 20.

PR #992, opened 2025-05-14
chore(deps): pin @ast-grep/napi 0.37.0 to keep support for ubuntu 20

Timeless0911:
## Checklist

### DIFF
--- a/packages/plugin-dts/package.json
+++ b/packages/plugin-dts/package.json
@@ -29,7 +29,7 @@
     "dev": "rslib build --watch"
   },
   "dependencies": {
-    "@ast-grep/napi": "^0.38.1",
+    "@ast-grep/napi": "0.37.0",
     "magic-string": "^0.30.17",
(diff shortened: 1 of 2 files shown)


Timeless0911:
@HerringtonDarkholme Can we keep support for ubuntu 20? 🫡

chenjiahan (approved):

HerringtonDarkholme:
unfortunately ast-grep is built by github action runner. GH recently deprecated ubuntu 20.

Timeless0911:
Cloudide in our company are still using Ubuntu 20 yet. And a glibc error will be throwed when the binding is loaded. And I'm still not clear about the situation of other docker images.
The same error in netlify with ubuntu 20 but we can update docker in Netlify.

Timeless0911:
img_v3_02m8_1420d220-2ff8-4a0f-99de-74a9741db36g

PR #998, opened 2025-05-15
feat: support `iife` format

fi3ework:
## Summary

Support `iife` format.

## Checklist

### DIFF
--- a/website/docs/en/guide/basic/output-format.mdx
+++ b/website/docs/en/guide/basic/output-format.mdx
@@ -140,14 +140,25 @@ MF stands for Module Federation. <MF />
 
 {/* The following documentation is taken from https://esbuild.github.io/api/#format-iife */}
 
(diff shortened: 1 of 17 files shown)


website/docs/en/guide/basic/output-format.mdx
```diff
@@ -135,3 +135,19 @@ export default defineConfig({
 ### What is MF?
 
 MF stands for Module Federation. <MF />
+
+## IIFE
+
+{/* The following documentation is taken from https://esbuild.github.io/api/#format-iife */}
+
+The iife format stands for "immediately-invoked function expression" and is intended to be run in the browser. Wrapping your code in a function expression ensures that any variables in your code don't accidentally conflict with variables in the global scope. If your entry point has exports that you want to expose as a global in the browser, you can configure that global's name using the global name setting. Specifying the `iife` format looks like this:
```
Timeless0911, line 143: > configure that global's name

It is esbuild option, we should rewrite it to our configurations?

fi3ework: yes we could do better by showing the globalObject option, i set it to globalThis here since it is more universal.

Timeless0911: Should we allow users to set `output.globalsObject` themselves (I think users config can override this though), or we should clearly state in the documentation that our default global name is `globalThis`?

Timeless0911 (approved):

PR #1056, opened 2025-06-14
chore(plugin-dts): remove error stack which is not helpful to users

Timeless0911:
## Summary

remove error stack which is not helpful to users.

### Before

image

image

### now

image

image

## Checklist


fi3ework:
Nice.

PR #1070, opened 2025-06-17
refactor: remove circular dependency and add plugin to detect

Timeless0911:
## Summary

- Remove existing circular dependency
image
- Enable `CircularDependencyRspackPlugin` to prevent importing circular dependency in the future

## Checklist


packages/core/src/css/const.ts
```diff
@@ -0,0 +1,6 @@
+export const BASE_URI = 'webpack://';
```
Timeless0911, line 1: If not extractor this file, `index.js` will output `import node_path, { extname as external_node_path_extname } from "node:path";`

Issue #1076, opened 2025-06-18
[Bug]: json import attribute not removed in bundleless mode


Netail:
### Version

```sh
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 574.64 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.12.0/bin/yarn
    npm: 10.9.0 - ~/.nvm/versions/node/v22.12.0/bin/npm
    bun: 1.2.10 - ~/.nvm/versions/node/v20.10.0/bin/bun
  Browsers:
    Chrome: 137.0.7151.105
    Edge: 137.0.3296.83
    Safari: 18.5
  npmPackages
    @rslib/core: 0.10.1
    @rsbuild/core: 1.4.0-beta.4
    @rspack/core: 1.4.0-beta.1
```

### Details

Import json is transformed to js, but the import statement import attributes (`with { type: 'json' }`) are not removed.

`import enUS from './en-US/translation.json' with { type: 'json' };` becoming `import translation from './en-US/translation.js' with { type: 'json' };`

### Reproduce link

-

### Reproduce Steps

Import a json file, e.g. `import enUS from './en-US/translation.json' with { type: 'json' };`. Build the library and see it being transformed into `import translation from './en-US/translation.js' with { type: 'json' };`, which is invalid

Timeless0911:
This is because in bundleless mode, the json file is added to the entry point by default.

The current bundleless implementation makes it very difficult to remove import attributes. If you want to use import attributes, the expected behavior in bundleless mode is to keep it.

Therefore, you can remove json files in `source.entry` and make json files externalized with `output.copy` to emit json files in output.

```ts
import { defineConfig } from '@rslib/core';

export default defineConfig({
  lib: [
    {
      bundle: false,
      source: {
        entry: {
          index: ['./src/**', '!./src/**/*.json'],
        },
      },
      output: {
        copy: [{ from: './**/*.json', context: './src' }],
        externals: [/.*\.json$/],
      },
    },
  ],
});
```

Netail:
Alright, bummer. Might be worth to mention in the docs :)

Timeless0911:
Great! I will update docs soon.

PR #1079, opened 2025-06-19
docs: use import attributes to import JSON files

Timeless0911:
## Summary

Add chapter of using import attributes to import JSON files.

## Related Links

close: https://github.com/web-infra-dev/rslib/issues/1076

## Checklist

### DIFF
--- a/website/docs/zh/guide/advanced/json-files.mdx
+++ b/website/docs/zh/guide/advanced/json-files.mdx
@@ -102,6 +102,16 @@ export { __webpack_exports__items as items, __webpack_exports__name as name };
 
 </Tabs>
 
+### 使用 import attributes
+
+在 bundle 模式下，Rslib 支持 [import attributes](https://github.com/tc39/proposal-import-attributes)，你可以通过 import attributes 来引入 JSON 文件：
+
+```js title="index.js"
+import json from './example.json' with { type: 'json' };
+```
+
+在 bundleless 模式下，通过 import attributes 引入 JSON 文件时，需要确保产物中保留对 JSON 文件的引用，参考 [文档](#bundleless) 进行配置。
+
 ## YAML 文件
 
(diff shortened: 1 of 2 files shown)


website/docs/zh/guide/advanced/json-files.mdx
```diff
@@ -102,6 +102,16 @@ export { __webpack_exports__items as items, __webpack_exports__name as name };
...
 
+### 使用 import attributes
+
+在 bundle 模式下，Rslib 支持 [import attributes](https://github.com/tc39/proposal-import-attributes)，你可以通过 import attributes 来引入 JSON 文件：
+
+```js title="index.js"
+import json from './example.json' with { type: 'json' };
+```
+
+在 bundleless 模式下，通过 import attributes 引入 JSON 文件时，需要确保产物中保留对 JSON 文件的引用，参考 [文档](#bundleless) 进行配置。
```
fi3ework, line 113: 1. It is not instructive enough, user still don't know what to do. Maybe attach the https://github.com/web-infra-dev/rslib/issues/1076 link is enough, or write the detail.
2. Is it ok to not write attributes in bundleless mode as a workaround?

Timeless0911: image
link jump to here
image
no need to write twice in one doc

Timeless0911: > Is it ok to not write attributes in bundleless mode as a workaround?

It is not possible to restrict the user's usage, and users may want to process json during secondary consumption

fi3ework: In bundleless mode, writing attributes does not affect the JSON that will be transformed into JavaScript. If the user wants to preserve the attributes in the output, they should include them in the source code, and vice versa.

PR #1090, opened 2025-06-26
test: use Rstest as Rslib's test framework

fi3ework:
## Summary

- [x] unit test
- [x] integration test
- [x] clean config
- [x] depend on latest version

Use Rstest as Rslib's test framework, with following shortage will be resolved in future versions:

- benchmark is disabled for now, despite it's not being used as of now.
- projects(workspace) is not supported yet, so some dep/devDep packages is effected in config snapshot.

## Checklist


packages/core/tests/__snapshots__/config.test.ts.snap
```diff
@@ -15,6 +15,16 @@ exports[`Should compose create Rsbuild config correctly > Merge Rsbuild config i
         },
         "externals": [
           [Function],
+          /\\^@rsbuild\\\\/core\\(\\$\\|\\\\/\\|\\\\\\\\\\)/,
```
Timeless0911, line 18: What's this?

packages/core/tests/__snapshots__/config.test.ts.snap
```diff
@@ -73,7 +83,7 @@ exports[`Should compose create Rsbuild config correctly > Merge Rsbuild config i
           "pnpapi",
         ],
         "filename": {
-          "js": "[name].mjs",
+          "js": "[name].js",
```
Timeless0911, line 86: why this changed

tests/integration/bundle-false/basic/src/dep.ts
```diff
@@ -1,3 +1,3 @@
-import { add } from 'dep_add';
+import depAdd from 'dep_add';
```
Timeless0911, line 1: What's the matter of named import?

tests/package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "type": "module",
   "scripts": {
+    "test": "pnpm rstest run",
```
Timeless0911:
```suggestion
    "test": "rstest run",
```

tests/package.json
```diff
@@ -3,6 +3,7 @@
...
     "test:benchmark": "vitest bench",
```
Timeless0911, line 7: vitest can be changed

Timeless0911:
```suggestion
    "test:benchmark": "rstest bench",
```

tests/rstest.config.ts
```diff
@@ -0,0 +1,18 @@
+// import codspeedPlugin from '@codspeed/vitest-plugin';
```
Timeless0911, line 1: We will have a `@codspeed/rstest-plugin` in the future?

website/docs/en/guide/advanced/module-federation.mdx
```diff
@@ -53,22 +53,25 @@ export default defineConfig({
...
+              react: {
+                singleton: true,
+              },
+              'react-dom': {
+                singleton: true,
+              },
             },
           },
-        }),
+          {},
```
Timeless0911, line 73: mf plugin add an extra required option?

fi3ework: yes, since 0.16.0.

tests/rstest.config.ts
fi3ework: we have to.

tests/integration/bundle-false/basic/src/dep.ts
fi3ework: image

Don't know why it passed in Vitest.

Timeless0911: Rollup support transforming commonjs to esm.

fi3ework: not related to Rollup, Vitest run all test in dev mode.

Timeless0911: I feel that this cjs can be statically analyzed and can be converted into an esm import using named import. Is this what webpack limited?

fi3ework: TBH, throwing an error complies node.js' behaviour. Does @9aoy got any clue?

packages/core/tests/__snapshots__/config.test.ts.snap
fi3ework: Because in the past, we're using Vitest project, somehow the cwd is not `packages/core`, and now we're running test under `packages/core` cwd, the deps and peerDeps then reads from `packages/core/package.json`, which is added here.

packages/core/tests/__snapshots__/config.test.ts.snap
fi3ework: Same as the above comment, root package.json is without `"type": "module"` whereas there is in `packages/core`.

tests/integration/bundle-false/basic/src/dep.ts
9aoy: This package has invalid named exports and relies on interop to run properly. Rstest implements this feature through proxy `import` syntax. reference: https://github.com/web-infra-dev/rstest/pull/331

image


However, rspack does not support parsing dynamic import expressions ☝️ , so rstest keeps the original dynamic import syntax here (reference: https://github.com/web-infra-dev/rstest/pull/231). Since `import` has no proxy, it will cause a runtime error.  

If [output.importFunctionName](https://rspack.rs/config/output#outputimportfunctionname) can take effect on dynamic import expression & `importDynamic: false`, maybe this case will work properly. @fi3ework any idea?

packages/core/tests/__snapshots__/config.test.ts.snap
9aoy: `pnpm -r test` to let each subproject execute rstest run separately, maybe it can temporarily implement similar effects to vitest workspace?

tests/integration/bundle-false/basic/src/dep.ts
9aoy: Update: this feature is not currently supported, because `import()` must contain at least some information about the location of the module in order to be bundled by rspack.

https://rspack.rs/api/runtime-api/module-methods#dynamic-expressions-in-import

packages/core/tests/__snapshots__/config.test.ts.snap
fi3ework: by coupled with monorepo?

Timeless0911 (dismissed):

fi3ework:
We need `hookTimeout` to let the test be stable.

PR #1106, opened 2025-07-07
fix: svgr should not work in css file

Timeless0911:
## Summary

svg files in css should not be transformed by svgr plugin in bundleless mode.

## Checklist


tests/integration/asset/svgr/rslib.config.ts
```diff
@@ -82,6 +97,11 @@ export default defineConfig({
     // cjs
     generateBundleCjsConfig({
       bundle: false,
+      source: {
+        entry: {
+          index: ['src', '!src/foo.css'],
```
SoonIter: NIT: could please rename `foo.css` to `css-entry.css` to explicitly indicate that this file is another test case

or else I cannot understand why you exclude this file

Timeless0911: ok

PR #1120, opened 2025-07-18
fix(dts): error log of set declarationDir when `dts.build` is true

Timeless0911:
## Summary

Should always set `declarationDir` if `outDir` or `declarationDir` not meet dts emit path.

## Checklist


Timeless0911:
forget to commit...

PR #1135, opened 2025-07-29
feat(dts): support `dts.alias`

Timeless0911:
### dts.alias

- **Type:**` Record<string, string>`
- **Default:** `{}`

Configure the path alias for declaration files.

`dts.alias` will be merged with `compilerOptions.paths` configured in `tsconfig.json` and `dts.alias` has a higher priority.

In most cases, you don't need to use `dts.alias`, but consider using it when you need to use path alias only in declaration files without wanting to affect JavaScript outputs. For example, map the declaration file of `foo` to `./compiled/foo`.

```ts title="rslib.config.ts"
export default {
  lib: [
    {
      // [!code highlight:5]
      dts: {
        alias: {
          foo: './compiled/foo',
        },
      },
    },
  ],
};
```

## Checklist


packages/plugin-dts/src/utils.ts
```diff
@@ -51,6 +51,33 @@ export function loadTsconfig(tsconfigPath: string): ts.ParsedCommandLine {
   return configFileContent;
 }
 
+export function mergeAliasWithTsConfigPaths(
```
fi3ework, line 54: Does this covered all cases? i'm not sure if `src/*` and `src/` can be merged correctly?

Timeless0911: `src/*` and `src/` are two different keys, as same as `foo` and `bar`.

fi3ework: early occurred key will be adopted by tsc.

PR #1159, opened 2025-08-12
feat!: update minimum Node.js version to 18.12.0

Timeless0911:
## Summary

Rspack and Rsbuild v1.5 will no longer support Node 16, as Node.js reached end-of-life on September 11th, 2023, and many npm packages in the ecosystem have also dropped support for Node 16.

This PR update the engines field in `package.json` to require Node.js >= 18.12.0.

## Related Links

https://github.com/web-infra-dev/rspack/discussions/10867

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   },
   "packageManager": "pnpm@10.14.0",
   "engines": {
-    "node": ">=18.0.0",
+    "node": ">=18.12.0",
     "pnpm": ">=10.14.0"
   }
 }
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
   },
   "packageManager": "pnpm@10.14.0",
   "engines": {
-    "node": ">=18.0.0",
+    "node": ">=18.12.0",
```
fi3ework, line 65: what specific feature we're relying on 18.12.0?

Timeless0911: align with upstream

fi3ework: make sense.

PR #1182, opened 2025-08-21
fix(css): prevent adding empty css rule

Timeless0911:
## Summary

`const rule = config.module.rule(ruleId);` will add an empty rule like below:

```
/* config.module.rule('sass') */
{},
/* config.module.rule('less') */
{},
/* config.module.rule('stylus') */
{},
```

When `@rsbuild/plugin-sass` begin to modify configurations, a unique id will be generated for Sass rules:

```
/* config.module.rule('sass-1') */
{},
```

Other plugins like [rsbuild-plugin-typed-css-modules](https://github.com/rspack-contrib/rsbuild-plugin-typed-css-modules) rely on `CHAIN_ID.RULE.SASS` to match rules to make features work, see https://github.com/web-infra-dev/rsbuild/issues/5909.

## Checklist

### DIFF
--- a/packages/core/src/css/cssConfig.ts
+++ b/packages/core/src/css/cssConfig.ts
@@ -96,6 +96,7 @@ const pluginLibCss = (
         CHAIN_ID.RULE.LESS,
         CHAIN_ID.RULE.STYLUS,
       ]) {
+        if (!config.module.rules.has(ruleId)) continue;
         const rule = config.module.rule(ruleId);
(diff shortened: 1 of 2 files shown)


Timeless0911:
Since we switch to bundleless build formly, so this is why css module types nolonger worked.

We can revert this PR https://github.com/web-infra-dev/rspress/pull/2476 when this fix is released in the near future. @SoonIter cc

fi3ework (approved):

PR #1193, opened 2025-08-27
fix: correct ESM interop

fi3ework:
## Summary

port https://github.com/web-infra-dev/rsbuild/pull/5955 before it can landed in Rsbuild (too much breaking as of now, may land it in v2).

fix https://github.com/web-infra-dev/rslib/discussions/1185.

## Checklist


Timeless0911 (approved):
Rsbuild configuration snapshot may need frequent updates, which remains to be seen.

PR #1196, opened 2025-08-28
fix: should glob dot files in bundleless mode

### LINKED ISSUE #1195: [Bug]: Hidden directories (starting with ".") are ignored when building in bundleless mode
### Version

```sh
System:
    OS: macOS 15.6.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 90.36 MB / 36.00 GB
    Shell: 4.0.2 - /opt/homebrew/bin/fish
  Browsers:
    Chrome: 139.0.7258.140
    Safari: 18.6
  npmPackages:
    @rslib/core: 0.12.3 => 0.12.3
```

### Details

When building in bundleless mode, directories whose names start with a dot (.) are ignored.
This causes files inside such directories not to be included in the build output.

### Expected behavior

Hidden directories (e.g., .generated) should be processed the same way as normal directories, unless explicitly excluded by configuration.

### Reproduce link

https://codesandbox.io/p/devbox/aged-currying-forked-xpkdq4?workspaceId=ws_4RzLomoT7g4GnfzMUfgpRN

### Reproduce Steps

1. Run the build process.

Karibash:
## Summary

Add an option to prevent glob from ignoring directories starting with a dot.

## Related Links

close #1195

## Checklist

### DIFF
--- a/tests/scripts/shared.ts
+++ b/tests/scripts/shared.ts
@@ -378,6 +378,9 @@ export function queryContent(
 ): { path: string; content: string } {
   const useBasename = options?.basename ?? false;
   const matched = Object.entries(contents).find(([key]) => {
+    if (key.includes('.hidden')) {
+      return false;
+    }
     const toQueried = useBasename ? basename(key) : key;
     return typeof query === 'string'
       ? toQueried === query
--- a/tests/scripts/shared.ts
+++ b/tests/scripts/shared.ts
@@ -378,9 +378,6 @@ export function queryContent(
 ): { path: string; content: string } {
   const useBasename = options?.basename ?? false;
   const matched = Object.entries(contents).find(([key]) => {
-    if (key.includes('.hidden')) {
-      return false;
-    }
     const toQueried = useBasename ? basename(key) : key;
     return typeof query === 'string'
       ? toQueried === query
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1348,7 +1348,11 @@ const composeBundlelessExternalConfig = (
                   // Requests that fall through here cannot be matched by any other externals config ahead.
                   // Treat all these requests as relative import of source code. Node.js won't add the
                   // leading './' to the relative path resolved by `path.relative`. So add manually it here.
-                  if (resolvedRequest[0] !== '.') {
+                  // should include dot files like '.hidden.ts'
+                  if (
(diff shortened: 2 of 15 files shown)


Timeless0911:
Thanks, can you add this case to tests `tests/integration/bundle-false/basic`?

Karibash:
It seems that when the file name contains a dot, an incorrect export is generated.
Does this require a fix in @rsbuild/core (or is it in @rspack/core instead)?

### Actual behavior

```js
export * from ".hiddenFile.js";
```

### Expected behavior

```js
export * from "./.hiddenFile.js";
```

Timeless0911:
This require a fix here that the logic of determine whether it is a relative path

https://github.com/web-infra-dev/rslib/blob/1bced60bb2b027c852735f7e00b96d82aa894e98/packages/core/src/config.ts#L1347-L1352

I think we can modify like this:

```ts
if (
  !resolvedRequest.startsWith('./') &&
  !resolvedRequest.startsWith('../')
) {
  resolvedRequest = `./${resolvedRequest}`;
}
```

Timeless0911:
Can I submit to this pr?

Karibash:
Thanks a lot!
I had no idea which part of the logic was causing the issue, so your explanation really helped me.

>Can I submit to this pr?

I think you’ll be able to submit it much faster than I could, so please go ahead and submit to the PR.

tests/scripts/shared.ts
```diff
@@ -378,6 +378,9 @@ export function queryContent(
 ): { path: string; content: string } {
   const useBasename = options?.basename ?? false;
   const matched = Object.entries(contents).find(([key]) => {
+    if (key.includes('.hidden')) {
+      return false;
+    }
```
fi3ework: why deliberately hide `.hidden`

Timeless0911: https://github.com/web-infra-dev/rslib/blob/1bced60bb2b027c852735f7e00b96d82aa894e98/tests/scripts/helper.ts#L17-L33

`globContentJSON` find path `index.js` when first matched. And `.hidden-folder/index.js` is before `index.js` in object. This is the fast way to modify less file.

Timeless0911: Oh I think i can modify the query

Timeless0911: done

fi3ework (approved):

PR #1198, opened 2025-08-28
fix: should add index main file when `redirect.js.path` set to `false`

Timeless0911:
## Summary

Align with `redirect.dts.path` behaviour

- When redirect.js.path is set to false, index should still be added before adding extension.
- When redirect.js.path is true, the resolver directly generate correct resolvedRequest with index appended.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1435,19 +1435,21 @@ const composeBundlelessExternalConfig = (
                     }
                   }
                 } else {
-                  // If the import path refers to a directory,
-                  // it most likely actually refers to a `index.*` file due to Node's module resolution.
-                  // When redirect.js.path is set to false, index should still be added before adding extension.
-                  // When redirect.js.path is true, the resolver directly generate correct resolvedRequest with index appended.
-                  if (
-                    !jsRedirectPath &&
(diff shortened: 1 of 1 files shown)


packages/core/src/config.ts
```diff
@@ -1434,6 +1435,17 @@ const composeBundlelessExternalConfig = (
                     }
                   }
                 } else {
+                  // If the import path refers to a directory,
+                  // it most likely actually refers to a `index.*` file due to Node's module resolution.
+                  // When redirect.js.path is set to false, index should still be added before adding extension.
+                  // When redirect.js.path is true, the resolver directly generate correct resolvedRequest with index appended.
+                  if (
+                    !jsRedirectPath &&
+                    (await isDirectory(join(dirname(issuer), resolvedRequest)))
```
fi3ework: i wonder if this will be a bit perf harmful as we will invoke this function for almost every file?

Timeless0911: Since the `redirect.js.path` default to true and rare users will set to `false`.

This logic only work when `redirect.js.path` set to `false` and `redirect.js.extension` set to `true`.

And `fsP.stat` seems not perf harmful?

fi3ework (approved):

PR #1220, opened 2025-09-12
fix: avoid chunk conflicts when multiple libs are present

fi3ework:
## Summary

to fix the case added in test cases, multi-compiler don't know other compilers' existence, so it's prone to emit chunk file to same path, whereas the content might not the same. this it's not prone to happen in Rsbuild as the hash is enabled but hash is disabled in Rslib.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -973,10 +973,11 @@ const composeExternalsConfig = (
   };
 };
 
-const composeAutoExtensionConfig = (
+const composeOutputFilenameConfig = (
   config: LibConfig,
   format: Format,
   autoExtension: boolean,
+  multiCompilerIndex: number | null,
   pkgJson?: PkgJson,
 ): {
   config: EnvironmentConfig;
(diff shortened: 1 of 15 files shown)


packages/core/src/config.ts
```diff
@@ -1093,6 +1093,25 @@ export const resolveEntryPath = (
...
+          output: {
+            chunkFilename: `${multiCompilerIndex}~[name].js`,
+          },
+        },
+      },
+    };
+  }
+
+  return {};
+};
```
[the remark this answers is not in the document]
fi3ework: good job.

packages/core/src/config.ts
```diff
@@ -1093,6 +1093,26 @@ export const resolveEntryPath = (
...
+  multiCompilerIndex: number | null,
+  format: Format,
+  jsExtension: string,
+): EnvironmentConfig => {
+  if (typeof multiCompilerIndex === 'number' && format !== 'mf') {
+    return {
+      tools: {
+        rspack: {
+          output: {
+            chunkFilename: `${multiCompilerIndex}~[name]${jsExtension}`,
```
Timeless0911: It should be inferred from `output.filename` in Rspack which can be also configured by `output.filename` and `output.filenameHash` in Rsbuild.

e.g. When `output.filenameHash` is set to true, the `output.chunkFilename` should be `${multiCompilerIndex}~[name].[contenthash:8]${jsExtension}`

packages/core/src/config.ts
```diff
@@ -1649,17 +1690,28 @@ async function composeLibRsbuildConfig(
...
+    pkgJson,
+  );
   const { entryConfig, outBase } = await composeEntryConfig(
     config.source?.entry!,
     config.bundle,
     rootPath,
     cssModulesAuto,
     config.outBase,
   );
+  // const outputConfig = composeOutputConfig(
```
Timeless0911: delete

PR #1232, opened 2025-09-17
chore: update tailwind v4 in test cases

Timeless0911:
## Summary

Update tailwind v4 in test cases.

## Related Links

https://tailwindcss.com/docs/installation

## Checklist


Timeless0911:
> should we reserve the current v3 tailwind test as v3 has 8M+ downloads while v4 has only 3M+ download as of now.

No longer necessary I think, Tailwind css V3 was not published in a year. And Rsbuild does not keep v3 tests too.

PR #1260, opened 2025-09-30
fix: treat `module` as normal variable in ESM

fi3ework:
## Summary

second (final) step after https://github.com/web-infra-dev/rspack/pull/11744.

forbid extra parser to handle `module` var if ESM syntax is present, common usage and user expectation in package building.

## Checklist

### DIFF
--- a/tests/integration/parser-javascript/api-plugin/src/cjsFile.cjs
+++ b/tests/integration/parser-javascript/api-plugin/src/cjsFile.cjs
@@ -1,7 +1,0 @@
-const value = () => 42;
-
-// Make the export immutable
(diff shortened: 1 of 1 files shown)


tests/integration/parser-javascript/api-plugin/src/cjsFile.cjs
```diff
@@ -0,0 +1,7 @@
+const value = () => 42;
```
Timeless0911: should be removed?

fi3ework:
@codex remove tests/integration/parser-javascript/api-plugin/src/cjsFile.cjs

PR #1280, opened 2025-10-21
fix: align esX browserslist expression to caniuse

fi3ework:
## Summary

previously, esX's browserslist expression came from esbuild's table, but esbuild's browser version was relative higher (for example, ES2015 supports dynamic import, so Chrome's minimum version was raised to `63` https://github.com/evanw/esbuild/blob/main/internal/compat/js_table.go#L470). 

here, we use data from caniuse to ensure a fix.

## Checklist


tests/integration/cli/build/build.test.ts
```diff
@@ -153,7 +153,7 @@ describe('build command', async () => {
     const files = await globContentJSON(path.join(fixturePath, 'dist'));
     expect(files).toMatchInlineSnapshot(`
       {
-        "<ROOT>/tests/integration/cli/build/no-config/dist/a/index.js": "const withoutConfig = 1000;
+        "<ROOT>/tests/integration/cli/build/no-config/dist/a/index.js": "var withoutConfig = 1000;
```
Timeless0911, line 156: seems not correct in personal experience.

PR #1331, opened 2025-11-12
fix: use native plugin to handle hashbang and react directives

fi3ework:
## Summary

fix https://github.com/web-infra-dev/rslib/issues/316.

## in this PR:

depends on depends on https://github.com/web-infra-dev/rspack/pull/12168, the native Rslib plugin has handled hashbang and react directives.

### with `advancedEsm`

no entry dedicated loader anymore, `advancedEsm` can handle "the issue of a module acting as both an entry module and a non-entry module is that the module graph can affect each other".

### without `advancedEsm`

still using a dedicated loader, but only an empty loader to make the doppelganger. the loader will not handle hashbang and react directives anymore (already handled in native plugin).

## Checklist

### DIFF
--- a/tests/integration/directive/react/bundleless/rslib.config.ts
+++ b/tests/integration/directive/react/bundleless/rslib.config.ts
@@ -12,7 +12,16 @@ export default defineConfig({
     generateBundleCjsConfig({
       bundle: false,
       output: {
-        minify: true,
+        minify: {
+          jsOptions: {
+            minimizerOptions: {
+              compress: {
+                // `directives` option is required to keep `"use strict"` in the output files.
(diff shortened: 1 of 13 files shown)


fi3ework:
@codex review

tests/integration/directive/react/bundleless/rslib.config.ts
```diff
@@ -12,7 +12,16 @@ export default defineConfig({
     generateBundleCjsConfig({
       bundle: false,
       output: {
-        minify: true,
+        minify: {
```
Timeless0911, line 15: remove this?

PR #1365, opened 2025-11-25
feat: default exclude `@types/react` from dts bundle

Timeless0911:
## Summary

- Api extractor can not bundle `@types/react`.

image

- A typical React component library often does not declare `@types/react` in `peerDependencies` but only in `devDependencies`. Following the autoExternal logic for dependency handling, Rslib will attempt to bundle `@types/react` into the declaration output files during the build. However, in practice, a component library should not bundle `@types/react`.

In this PR, default exclude `@types/react` from dts bundle.

## Checklist

### DIFF
--- a/packages/plugin-dts/src/dts.ts
+++ b/packages/plugin-dts/src/dts.ts
@@ -20,6 +20,8 @@ import {
 const isObject = (obj: unknown): obj is Record<string, any> =>
   Object.prototype.toString.call(obj) === '[object Object]';
 
+export const DEFAULT_EXCLUDED_PACKAGES: string[] = ['@types/react'];
+
 // use !externals
 export const calcBundledPackages = (options: {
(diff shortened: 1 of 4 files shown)


fi3ework (commented):
in terms of the best practice, do we need to add `@types/react` to peer deps in our template?

Timeless0911:
> in terms of the best practice, do we need to add `@types/react` to peer deps in our template?

In fact, almost all tool templates don't include `@types/react` and `@types/react-dom` in the peer deps. This is likely because React versions within the monorepo projects in community are usually quite consistent, avoiding the type conflicts caused by pnpm's hoisting behavior. 

I prefer to temporarily not add them to peer deps in the templates, and we just handle `@types/react` to prevent it from being bundled, especially since api-extractor can not bundle `@types/react` though.

PR #1386, opened 2025-12-09
feat: improve bundleless file size print

Timeless0911:
### Before

image

### Now

image

## Checklist


fi3ework:
- it would be better if the `total` could be in same style.
image

- and i wonder why there's no gzipped size for esm2

Timeless0911:
> and i wonder why there's no gzipped size for esm2

Only web target will dispaly gzipped size

PR #1397, opened 2025-12-17
fix(plugin-dts): follow `lib.dts.abortOnError` when generating dts

zthxxx:
## Summary

- pass `abortOnError` through dts pipeline (tsc/tsgo) instead of always failing when type-check has error
- allow bundling to proceed when `abortOnError` is `false` despite diagnostics

## Related Links

N/A

## Checklist

- [x] Tests updated (not required).
- [x] Documentation updated (not required).

### DIFF
--- a/packages/plugin-dts/src/tsc.ts
+++ b/packages/plugin-dts/src/tsc.ts
@@ -28,6 +28,7 @@ const formatHost: FormatDiagnosticsHost = {
 };
 
 export type EmitDtsOptions = {
+  abortOnError: boolean | undefined;
   name: string;
   cwd: string;
   configPath: string;
@@ -42,6 +43,7 @@ export type EmitDtsOptions = {
 };
 
 async function handleDiagnosticsAndProcessFiles(
+  abortOnError: boolean | undefined,
   diagnostics: readonly Diagnostic[],
   configPath: string,
   bundle: boolean,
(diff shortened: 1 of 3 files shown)


packages/plugin-dts/src/tsc.ts
```diff
@@ -81,12 +83,14 @@ async function handleDiagnosticsAndProcessFiles(
       logger.error(logPrefixTsc, message);
     }
 
-    const error = new Error(
-      `Failed to generate declaration files. ${color.dim(`(${name})`)}`,
-    );
-    // do not log the stack trace, diagnostic messages are enough
-    error.stack = '';
-    throw error;
+    if (abortOnError) {
```
Timeless0911, line 86: When `bundle: false` with `abortOnError: false`, the error should still be throwed I think. `abortOnError` means not exit the process instead of not throwing error.

zthxxx: If we throw an error here, the outer dts generation flow will fail due to promise rejection, no matter whether `bundle` is `true` or `false`.

`packages/plugin-dts/src/dts.ts:265`

In other words, even when `abortOnError` is set to false, throwing here will still cause the whole external pipeline to stop and skip subsequent steps.

That’s why the error is handled without throwing in this case, allowing the process to continue as intended when `abortOnError` is disabled.

Timeless0911: Promise rejection is handled in main process. https://github.com/web-infra-dev/rslib/blob/22abbc9ee52aa9a21f515d3dab065d5e2380c218/packages/plugin-dts/src/index.ts#L238-L252

When `abortOnError` is set to false, a warning will be printed and the process continue.

image

Timeless0911: We may optimize the behaviour when `abortOnError` set to false with `bundle` set to true. Now, if tsc/tsgo throws error, the logic in `apiExtractor.ts` will not be called.

When `bundle` set to false, throwing error is reasonable since no other operations will be performed.

In other words, `abortOnError` is used to prevent a process from exiting with a non-zero exit code. If you want to ignore type errors, you should set `noCheck` to disable type checking.

zthxxx: Got it, `noCheck` is which I need

PR #1572, opened 2026-04-02
feat(core): add `experiments.exe` support

Timeless0911:
## Summary

Add experimental `experiments.exe` support so Rslib can package bundled Node.js outputs as single executable applications.

The new flow resolves target binaries, handles cross-platform target generation, and exposes configuration for file naming, output paths, targets, and SEA options.

### Basic API shape

``` ts
type ExeOptions =
  | boolean
  | {
      fileName?: string;
      outputPath?: string;
      targets?: Array<
        | string
        | {
            platform?: 'darwin' | 'linux' | 'win32';
            arch?: 'x64' | 'arm64';
            nodeVersion?: string;
          }
      >;
      seaOptions?: SeaOptions;
    };
```

### Basic usage
```ts
export default {
  lib: [
    {
      format: 'esm',
      experiments: {
        exe: {
          fileName: 'hello',
        },
      },
    },
  ],
};
```
## Related Links

- Node.js SEA documentation: https://nodejs.org/api/single-executable-applications.html

## Credits

Rslib's SEA bundling support was inspired by prior work in [tsdown](https://github.com/rolldown/tsdown), created by [Kevin Deng](https://github.com/sxzz).

## Checklist


packages/core/src/exe/utils.ts
```diff
@@ -0,0 +1,189 @@
+import { spawn } from 'node:child_process';
+import path from 'node:path';
+import type { EnvironmentContext, Rspack } from '@rsbuild/core';
+import { color } from '../utils/color';
+import type { CommandResult, NormalizedExeTarget } from './types';
+
+export const isJavaScriptOutputFile = (file: string): boolean => {
+  return /\.(cjs|mjs|js|jsx)$/.test(file) && !file.endsWith('.map');
```
chenjiahan: `!file.endsWith('.map')` seems to be unnecessary

packages/core/src/exe/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+};
+
+export const formatBinarySize = (size: number): string => {
+  if (size >= 1024 * 1024) {
+    return `${(size / (1024 * 1024)).toFixed(1)} MB`;
+  }
+
+  if (size >= 1024) {
+    return `${(size / 1024).toFixed(1)} kB`;
```
chenjiahan: 1 kB = 1000 B, not 1024

1 KiB = 1024 B

Timeless0911: No agent noticed this during the AI ​​code review.😂

website/docs/zh/config/lib/experiments.mdx
```diff
@@ -46,3 +46,267 @@ export default {
...
+      >;
+      seaOptions?: SeaOptions;
+    };
+```
+
+- **默认值：** `false`
+
+::: tip Node.js 要求
+
+`experiments.exe` 要求 Node.js 25.7.0 或更高版本，不支持在 Bun 或 Deno 中使用。
```
fi3ework: what's our plan to support Deno and Bun SEA in the future, or we only need to add a simple integration with their native SEA support?

Timeless0911: Let's wait until deno and bun support this feature.

- https://bun.com/reference/node/sea
- https://docs.deno.com/api/node/sea

fi3ework: i think there are two issues here: for short-to-medium-term users, if they are building an exe with Bun/Deno, they can use Rslib to directly bundle a product, and then use `bun build --compile` on this bundle file to generate the exe (similarly for Deno).

i personally don't think there's much point in using Bun/Deno's Node flavor SEA.

so what I want to express is that the wording here should be changed to: if using Bun/Deno, they can directly execute commands like `bun build --compile` on Rslib's bundle output.

Timeless0911: done

packages/core/src/exe/download.ts
```diff
@@ -0,0 +1,393 @@
...
+    platform === 'win32'
+      ? path.join(extractDir, 'node.exe')
+      : path.join(extractDir, 'bin', 'node');
+
+  return {
+    archiveName,
+    archivePath,
+    binaryPath,
+    extractDir,
+    url: `https://nodejs.org/dist/${normalizedVersion}/${archiveName}`,
```
fi3ework, line 69: maybe we can support custom mirror registry in the future, for users use domestic mirror.

fi3ework (approved):

packages/core/src/exe/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+import path from 'node:path';
+import type { EnvironmentContext, Rspack } from '@rsbuild/core';
+import { color } from '../utils/color';
+import type { CommandResult, NormalizedExeTarget } from './types';
+
+export const isJavaScriptOutputFile = (file: string): boolean => {
+  return /\.(cjs|mjs|js|jsx)$/.test(file);
+};
+
+export const formatBinarySize = (size: number): string => {
```
fi3ework, line 11: too much code, what if use https://github.com/sindresorhus/pretty-bytes and https://github.com/sindresorhus/pretty-ms

fi3ework (commented):
.

packages/core/src/exe/utils.ts
Timeless0911: The bundled output will be larger if so.

fi3ework: not likely, because we only bundle the code that's needed. moreover, compared to manually maintaining this code (including edge cases), i think the trade-off of using it directly is very acceptable.

fi3ework: i think there's another reason behind it, because this code is written by the agent, so it's capable of writing it. if we were writing it ourselves, we would naturally tend to use existing npm packages, especially since these packages are well maintained by sindresorhus. 

i don't know which way is right, but intuitively I want to maintain less code, especially since this code basically doesn't need to be changed anymore, which can also reduce the disclosure of context to the agent.

fi3ework (approved):

PR #1659, opened 2026-05-20
docs: fix stale Rsbuild doc links

elecmonkey:
## Summary

This PR cleans up stale Rsbuild links in the Rslib docs.

- Fix links that now lead to 40x pages
- Replace links that return 30x redirects with their final URLs
- zh-CN docs should link to zh-CN docs


chenjiahan (approved):
LGTM 👍

PR #1672, opened 2026-05-26
feat(core)!: use modern-module externals type for ESM

Timeless0911:
## Summary

This PR changes the default ESM `externalsType` to Rspack's `modern-module` for all output targets. Static `import` externals remain ESM imports, dynamic `import()` externals remain dynamic, and CommonJS `require()` externals use `createRequire` for Node-like targets or a bare `require()` for other targets.

It also removes the legacy CommonJS external warning path and the no-longer-needed `autoCjsNodeBuiltin` RslibPlugin option.

When external modules loaded through `createRequire()` are bundled again, the consuming bundler must statically analyze this call pattern; Rspack supports it through `module.parser.javascript.createRequire`.

## Related Links

https://rspack.rs/config/externals#externalstypemodern-module
https://rspack.rs/config/module-parser#javascriptcreaterequire

## Checklist


Timeless0911:
pending
1. Rspack support bundle `createRequire`
2. `modern-module` external type now depends `externalsPreset.node` to determine node target, but Rslib set `externalsPreset.node` to `false` since https://github.com/web-infra-dev/rslib/pull/1318 due to `externals: { fs: false }` can not takes effect with externalsPreset which should be a bug of Rspack externals?

PR #1703, opened 2026-06-16
perf(core): use deterministic module ids for web targets

Timeless0911:
## Summary

This PR optimizes web-targeted bundled output by defaulting non-MF module ids to deterministic ids instead of named ids, reducing path-like module id strings in production artifacts while preserving named ids for node targets and MF dev. It also centralizes the default module id composition and keeps user `tools.rspack.optimization.moduleIds` overrides working.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -482,7 +482,6 @@ export async function createConstantRsbuildConfig(): Promise<EnvironmentConfig> 
       htmlPlugin: false,
       rspack: {
         optimization: {
-          moduleIds: 'named',
           nodeEnv: false,
         },
         // TypeScript-specific behavior: if the extension is ".js" or ".jsx", try replacing it with ".ts" or ".tsx"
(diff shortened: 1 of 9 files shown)


packages/core/src/config.ts
```diff
@@ -1705,6 +1703,34 @@ const composeTargetConfig = (
...
+  format: Format,
+  target: RsbuildConfigOutputTarget,
+): EnvironmentConfig => {
+  if (format === 'mf') {
+    return {
+      tools: {
+        rspack: (config, { env }) => {
+          config.optimization = {
+            ...config.optimization,
+            moduleIds: env === 'development' ? 'named' : 'deterministic',
```
chenjiahan: Can we just use the Rspack default value?

packages/core/tests/config.test.ts
```diff
@@ -737,6 +737,74 @@ describe('syntax', () => {
   });
 });
 
+describe('module ids', () => {
+  test('uses named module ids for node target by default', async () => {
+    const rslibConfig: RslibConfig = {
+      lib: [
+        {
+          format: 'esm',
```
chenjiahan: `format: 'esm'` is the default value. Can we remove this?

chenjiahan (approved):

PR #1707, opened 2026-06-17
fix(core): preserve query in bundleless redirects

Timeless0911:
## Summary

This PR fixes bundleless redirect handling for JS and style imports with query strings. JS/TS redirects now preserve query and fragment parts while rewriting extensions, plain CSS keeps query strings when redirecting to emitted CSS files, and CSS Modules with query strings redirect to the emitted CSS asset instead of putting the query on the JS wrapper. Asset and SVGR query semantics are intentionally left out of this PR for a separate bundleless asset output design.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -63,6 +63,11 @@ import {
   pick,
   readPackageJson,
 } from './utils/helper';
+import {
+  appendPathExtension,
+  parsePathQueryFragment,
+  replacePathExtension,
+} from './utils/path';
 import { isDebug, logger } from './utils/logger';
 import {
   ESX_TO_BROWSERSLIST,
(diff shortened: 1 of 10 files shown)


Timeless0911:
### Current issue and scope of this PR

The core issue is that bundleless redirect only rewrites import specifiers, while artifact generation is handled separately. Simply preserving a query does not mean that the semantics of `?raw`, `?inline`, or `?react` have actually been implemented.

This draft PR is currently scoped as follows:

| Type | Current behavior |
| --- | --- |
| JavaScript | Preserve the query/fragment while rewriting the extension as usual, e.g. `foo.ts?x` → `foo.js?x` |
| Plain CSS | Preserve the query and redirect to the extracted CSS file, e.g. `index.less?inline` → `index.css?inline` |
| CSS Modules | Without a query, continue redirecting to the JS wrapper; with a query, redirect to the extracted CSS file, e.g. `index.module.less?inline` → `index_module.css?inline` |
| Regular assets such as PNG | Keep the existing behavior and leave query support out of this PR |
| SVGR | Leave it out of this PR to avoid preserving a query on an unsupported query import when no corresponding wrapper is emitted |

The CSS Modules handling covers different preprocessors rather than only `.module.css`:

```text
index.module.less  -> index_module.css
index.module.scss  -> index_module.css
index.module.styl  -> index_module.css
```

Regular assets cannot use the same path-preservation logic directly because bundleless asset handling currently looks roughly like this:

```text
source PNG
  -> emit into static
  -> generate an intermediate JS wrapper
  -> application code imports the JS wrapper
```

For example, redirect could preserve this import:

```js
import raw from './logo.png?raw';
```

However, if the build only emits a JS wrapper or `static/logo.png`, there is no artifact corresponding to the preserved `./logo.png?raw` specifier. SVGR imports such as `?react` and `?url` have the same underlying issue, with the additional question of whether they should emit a React wrapper, a URL wrapper, or externalize/copy the original file.

Therefore, this PR only fixes cases where JS, CSS, and CSS Modules already have a well-defined output artifact and redirect was dropping the query. Asset and SVGR support should be designed separately together with asset externalization/copying and proxy-module generation. Bundle mode continues to rely on bundler/loader query handling; this change is focused on bundleless redirect behavior and is not intended to change bundle-mode semantics.

PR #1710, opened 2026-06-22
fix(core): prefix runtime chunk in multi-compiler builds

Timeless0911:
## Summary

Fixes runtime chunk naming for bundled ESM multi-compiler builds by explicitly prefixing runtime chunk names with the compiler index even when an individual compiler has one or default entry. This prevents multiple compilers from falling back to the same default runtime chunk name while preserving single-compiler behavior.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -835,7 +835,11 @@ export const getRuntimeChunkConfig = ({
     };
   }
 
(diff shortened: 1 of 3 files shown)


JSerFeng (approved):

Timeless0911:
This affects the optimization of inline runtime chunks.

PR #1734, opened 2026-06-30
feat!: update `lib.syntax` browserslist baselines

elecmonkey:
## BREAKING CHANGE

`lib.syntax` now uses explicit browserslist baselines for ES2023, ES2024, and ES2025.

**This should have limited practical impact**. 

- JavaScript downleveling is mostly unchanged because Rspack & SWC's minifier currently clamp modern ES targets to the ES2022 transform baseline. SWC's only wired post-ES2022 syntax transform is the RegExp v flag, and both the old and new baselines land on the same side.
- The main difference is that Lightning CSS may emit more modern CSS output for these syntax targets. Use `esnext` to keep the dynamic latest-browser/latest-Node behavior.

## Summary

This PR adds `es2023`, `es2024`, and `es2025` as fixed `syntax` values and gives each its own browserslist baseline.

Mapping every newer syntax value to an ES2022-era browser floor is too conservative for downstream browserslist consumers: it can pin Lightning CSS to an older baseline even when the user targets a newer ECMAScript version, so modern CSS features can get downleveled unnecessarily.

In Rslib, `syntax` drives two independent paths:

- `config.target` (a literal ES version) -> Rspack runtime codegen and target-derived defaults.
- `output.overrideBrowserslist` -> the SWC loader's `env.targets` (user-code transforms) and Lightning CSS.

So the browserslist table directly controls which syntax SWC downlevels and which CSS features Lightning CSS emits.

## Before This PR

The pre-PR `main` branch only had explicit support through `es2024`.

| syntax | pre-PR browserslist behavior | chrome/edge | firefox | safari/ios | node |
|---|---|---:|---:|---:|---:|
| `es2023` | reused `es2022` | 94 | 93 | 16.4 | 16.11 |
| `es2024` | reused `esnext` dynamic query | target-dependent | target-dependent | target-dependent | target-dependent |

The old `es2024` value was not a fixed browser table. It reused the dynamic `esnext` query:

| output target | pre-PR `es2024` / `esnext` query |
|---|---|
| `node` | `last 1 node versions` |
| `web` / `web-worker` | `last 1 Chrome versions`, `last 1 Firefox versions`, `last 1 Edge versions`, `last 1 Safari versions`, `last 1 ios_saf versions`, `not dead` |

## New Fixed Baselines

These values are not intended to be an exact ECMAScript-spec support matrix. They are Rslib compatibility baselines used for both SWC JavaScript transforms and Lightning CSS.

The current rule is:

- Keep `es2023` modern enough for Lightning CSS, but below the RegExp `v` flag support line so ES2024 syntax is still downleveled when users select `syntax: 'es2023'`.
- Keep `es2024` at or above SWC's currently wired post-ES2022 transform: RegExp `v` flag / `UnicodeSetsRegex` (`transform-unicode-sets-regex`).
- Keep `es2025` at a newer fixed baseline based on the later regexp features in Babel/SWC compat-data.
- Keep `esnext` as the dynamic latest-browser/latest-Node query.

| syntax | chrome/edge | firefox | safari/ios | node | rationale |
|---|---:|---:|---:|---:|---|
| `es2023` | 110 | 115 | 17 | 20 | modern CSS/browser baseline, intentionally below the RegExp `v` flag support line |
| `es2024` | 112 | 116 | 17 | 20 | `transform-unicode-sets-regex` baseline for RegExp `v` flag |
| `es2025` | 126 | 132 | 17.4 | 23 | fixed modern baseline based on later regexp compat-data |

Notes:

- `es2023` and `es2024` intentionally do **not** share the same baseline. The RegExp `v` flag is an ES2024 feature, so `syntax: 'es2023'` should not make SWC preserve it.
- `es2022` and below are unchanged.
- `esnext` remains the dynamic "latest" query.


packages/core/src/utils/syntax.ts
Timeless0911: You should modify this mapping.

Timeless0911: and the `packages/core/src/config.ts:1120` still be  `config.target = ['es2022']`. This pr should be a fix of esVersions mappings and more support of esversion lists.

packages/core/src/utils/syntax.ts
```diff
@@ -140,14 +151,12 @@ export function transformSyntaxToRspackTarget(
 
     if (normalizedSyntaxItem.startsWith('es')) {
       if (normalizedSyntaxItem in ESX_TO_BROWSERSLIST) {
-        // The latest EcmaScript version supported by Rspack's `target` is es2022.
-        // Higher versions are treated as es2022.
         if (
           RSPACK_TARGET_UNLISTED_MODERN_ECMA_VERSIONS.includes(
```
Timeless0911, line 165: `RSPACK_TARGET_UNLISTED_MODERN_ECMA_VERSIONS` now includes `es2025`, but the fallback also returns `es2025`. Since `es2025` is a supported concrete Rspack target, it no longer fits the “unlisted” concept.

packages/core/tests/syntax.test.ts
```diff
@@ -112,8 +113,8 @@ describe('transformSyntaxToBrowserslist', () => {
         "last 1 node versions",
       ]
     `);
```
Timeless0911, line 180: Since this PR adds explicit fixed browserslist mappings for `es2023` and `es2024`, can we add targeted tests for the concrete outputs of `transformSyntaxToBrowserslist('es2023', 'web')` and `transformSyntaxToBrowserslist('es2024', 'web')`?

packages/core/tests/syntax.test.ts
```diff
@@ -28,9 +28,37 @@ const compareSemver = (a: string, b: string) => {
...
+    for (const version of ['es2024', 'es2025'] as const) {
+      const baseline = ESX_TO_BROWSERSLIST[version] as Record<string, string>;
+      for (const [engine, min] of Object.entries(unicodeSetsRegex)) {
+        expect(compareSemver(baseline[engine]!, min)).toBeGreaterThanOrEqual(0);
+      }
+    }
+  });
+
+  test('es2023 should stay below the RegExp v flag support line', () => {
+    expect(
```
Timeless0911, line 53: we can add a exact test expect(ESX_TO_BROWSERSLIST.es2023).toStrictEqual({
  chrome: '110',
  edge: '110',
  firefox: '115',
  ios: '17',
  node: '20',
  safari: '17',
});

Timeless0911: oh i see the below tests

Timeless0911 (approved):

PR #1735, opened 2026-07-01
fix(core): reset Rspack const environment

Timeless0911:
## Summary

This PR upgrades Rsbuild to 2.1.2 and resets Rsbuild's web-like output `environment.const` default for Rslib library builds. Rslib now removes the app-runtime `const: false` override so Rspack can infer `const` support from the configured target, while explicit user overrides through `tools.rspack` and lib-level plugins remain supported.

It also consolidates Rslib's format-level Rsbuild default adjustments into `rslib:modify-rsbuild-default`, registered from `composeFormatConfig`:

- disables `new URL()` parsing for ESM/CJS library output
- removes Rsbuild's `type: 'javascript/auto'` override so ESM-like modules follow Rspack's original inference and are treated as strict ESM (`javascript/esm`)
- resets the web/web-worker `environment.const` override

## Related Links

https://github.com/web-infra-dev/rspack/releases/tag/v2.1.2
https://github.com/web-infra-dev/rsbuild/pull/8020

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -761,26 +761,36 @@ const composeFormatConfig = ({
   }
 };
 
-const disableUrlParseRsbuildPlugin = (): RsbuildPlugin => ({
-  name: 'rsbuild:disable-url-parse',
+const modifyRsbuildDefaultPlugin = ({
+  disableUrlParse,
+}: {
+  disableUrlParse?: boolean;
+} = {}): RsbuildPlugin => ({
+  name: 'rslib:modify-rsbuild-default',
   setup(api) {
-    api.modifyBundlerChain((config, { CHAIN_ID }) => {
+    api.modifyBundlerChain((chain, { CHAIN_ID, target }) => {
+      // Part 1: disable URL parsing for library output.
       // Fix for https://github.com/web-infra-dev/rslib/issues/499.
-      // Prevent parsing and try bundling `new URL()` in ESM format.
-      config.module
+      // Prevent parsing and try bundling `new URL()` in ESM/CJS format.
+      if (disableUrlParse) {
+        chain.module
+          .rule(CHAIN_ID.RULE.JS)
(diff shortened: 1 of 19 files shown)


packages/core/src/config.ts
```diff
@@ -777,6 +777,31 @@ const disableUrlParseRsbuildPlugin = (): RsbuildPlugin => ({
...
+
+      const environment = chain.output.get('environment');
+
+      if (!environment || environment.const !== false) {
+        return;
+      }
+
+      // Rsbuild disables `const` for web-like app runtimes. Rslib should
+      // leave this to Rspack's target inference for library output.
+      delete environment.const;
```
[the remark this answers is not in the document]
Timeless0911: Currently by design, global plugins can not affect internal environment plugin, users can directly set `tools.rspack` in global to override.

packages/core/src/config.ts
```diff
@@ -777,6 +777,31 @@ const disableUrlParseRsbuildPlugin = (): RsbuildPlugin => ({
   },
 });
 
+const resetEnvConstPlugin = (): RsbuildPlugin => ({
```
chenjiahan: Can we merge these small Rsbuild plugins to a unified plugin?

packages/core/src/config.ts
```diff
@@ -761,26 +761,36 @@ const composeFormatConfig = ({
...
+        chain.module
+          .rule(CHAIN_ID.RULE.JS)
+          .oneOf(CHAIN_ID.ONE_OF.JS_MAIN)
+          .parser({
+            url: false,
+          });
+      }
+
+      // Part 2: remove Rsbuild's default JS module type.
+      // Port https://github.com/web-infra-dev/rsbuild/pull/5955 before it merged into Rsbuild.
```
chenjiahan: Is this comment outdated?

chenjiahan (approved):

PR #1739, opened 2026-07-03
feat(create-rslib): bump default Node syntax to 22 and add public publishConfig

elecmonkey:
## Summary

- **Bump the default Node syntax from `node 18` to `node 22`** (Node 18/20 are EOL) in the `node-esm` and `node-dual` templates.
- **Drop redundant `format: 'esm'`** since Rslib defaults `lib.format` to `'esm'`. `node-dual` keeps `esm` + `cjs` to show the dual-output intent, and `svelte` keeps it because its `lib` entry has no other fields.
- **Add `publishConfig.access: "public"` to every base template** so scoped packages can be published without extra setup (with a matching assertion in `test/helper.ts`).

### DIFF
--- a/packages/create-rslib/template-node-dual-js/package.json
+++ b/packages/create-rslib/template-node-dual-js/package.json
@@ -12,6 +12,9 @@
   "files": [
     "dist"
   ],
+  "publishConfig": {
+    "access": "public"
+  },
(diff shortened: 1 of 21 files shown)


chenjiahan (approved):
LGTM 👍

PR #1747, opened 2026-07-07
refactor(dts): avoid duplicated error label in diagnostics

Timeless0911:
## Summary

This PR removes the duplicated `error` word from TypeScript diagnostics emitted by the dts TS API backend. It keeps the outer Rslib `error` log level while preserving TypeScript's colored file location, diagnostic code, and code frame output.

### Before

image

### After
image

## Checklist

### DIFF
--- a/packages/plugin-dts/src/tsc.ts
+++ b/packages/plugin-dts/src/tsc.ts
@@ -20,6 +20,21 @@ import {
 } from './utils';
 
 const logPrefixTsc = color.dim('[tsc]');
+/*
+ * TS builds this prefix in `formatDiagnosticsWithColorAndContext`:
+ * `diagnosticCategoryName` + `getCategoryFormat` + `formatColorAndReset`.
+ * Rslib already prints `logger.error`, so strip only the TS category word.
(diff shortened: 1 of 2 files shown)


packages/plugin-dts/src/tsc.ts
```diff
@@ -20,6 +20,18 @@ import {
...
 const logPrefixTsc = color.dim('[tsc]');
+/*
+ * TS builds this prefix in `formatDiagnosticsWithColorAndContext`:
+ * `diagnosticCategoryName` + `getCategoryFormat` + `formatColorAndReset`.
+ * Rslib already prints `logger.error`, so strip only the TS category word.
+ * Use String.fromCharCode(27) for ESC to avoid no-control-regex.
+ * See https://github.com/microsoft/TypeScript/blob/main/src/compiler/program.ts
+ */
+const ansiEscape = String.fromCharCode(27);
+const diagnosticErrorPrefixRE = new RegExp(
```
chenjiahan: Can we create the regex lazily?

Timeless0911: done

chenjiahan (approved):

PR #1750, opened 2026-07-07
fix(create-rslib): move reference types into tsconfig

chenjiahan:
## Summary

This PR moves create-rslib template type references from triple-slash `.d.ts` files into `compilerOptions.types` in generated `tsconfig.json` files.

## Checklist

### DIFF
--- a/packages/create-rslib/template-react-ts/tests/tsconfig.json
+++ b/packages/create-rslib/template-react-ts/tests/tsconfig.json
@@ -1,4 +1,7 @@
 {
   "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "types": ["@rslib/core/types", "@testing-library/jest-dom"]
+  },
   "include": [".", "../rstest.setup.ts"]
 }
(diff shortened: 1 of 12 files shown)


packages/create-rslib/template-react-ts/tests/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "types": ["@rslib/core/types", "@testing-library/jest-dom"]
```
Timeless0911, line 4: Do we still need the rslib type in the tsconfig file that belongs to the testing section?

chenjiahan: Yes, because `compilerOptions.types` is not merged from the extended config. The test files also import source files from `src`, so once the test tsconfig declares `@testing-library/jest-dom`, it needs to include `@rslib/core/types` as well.

Timeless0911: Got it.

Timeless0911 (approved):

PR #1755, opened 2026-07-08
feat: support shared lib config fields

elecmonkey:
## Summary

This PR makes the already-supported shared `lib` configuration pattern part of the public TypeScript API. Rslib can merge selected top-level library options into each `lib` item at runtime, and this change exposes that behavior in `RslibConfig` types for fields that are safe to use as defaults: `bundle`, `autoExtension`, `autoExternal`, `redirect`, `syntax`, `externalHelpers`, `banner`, `footer`, `shims`, and `outBase`.

Options that identify or materially specialize a single output, such as `format`, `id`, `dts`, `umdName`, and `experiments`, remain per-`lib` only.

## Checklist


packages/core/src/types/config.ts
```diff
@@ -356,6 +356,13 @@ export type Redirect = {
 };
 
 export type LibExperiments = {
+  /**
+   * Whether to enable Rspack advanced ESM output.
+   * @deprecated The advanced ESM output is now the default output for ESM format and this option has no effect. It will be removed in a future major version.
+   * @defaultValue `true`
+   * @see {@link https://rslib.rs/config/lib/experiments#experimentsadvancedesm}
+   */
+  advancedEsm?: boolean;
```
Timeless0911: this has been removed

Timeless0911 (approved):

PR #1756, opened 2026-07-08
feat(core): allow omitting lib config

elecmonkey:
## Summary

This PR allows `lib` to be omitted from Rslib config when the project only needs the default single library output. Omitting `lib` now has the same behavior as `lib: [{}]`, while explicit invalid values such as `lib: []` or `lib: null` still fail validation instead of being treated as defaults.

The CLI option path also initializes the implicit default `lib` item so flags such as `--format`, `--entry`, and `--dts` continue to apply when `lib` is omitted.

## Checklist


Timeless0911:
Check [mergeRslibConfig](https://rslib.rs/zh/api/javascript-api/core#mergerslibconfig) and modify some docs or anything need to be changed.

Timeless0911:
@codex review

Timeless0911 (approved):

PR #1760, opened 2026-07-09
refactor(dts)!: remove native-preview support

Timeless0911:
## Summary

This PR removes the deprecated `@typescript/native-preview` support path for declaration generation. `dts.tsgo` now resolves the native TypeScript executable only from `typescript` >= 7.0.0, and explicit `dts.tsgo: true` without TypeScript 7+ now fails with a clear error. This is a breaking change for projects relying on `@typescript/native-preview`; install TypeScript 7+ instead.

The docs are updated to point at `typescript@latest`.

## TODO

- Update examples, templates, and website to `typescript@7` after upstream packages support the TypeScript 7 peer range.
- Keep `prebundle` on root `typescript@^6.0.3` for now, because its toolchain still depends on the TypeScript 6-compatible peer/API surface.

## Checklist


Timeless0911:
@codex review

fi3ework (approved):

PR #1769, opened 2026-07-17
feat(core): support reload-server watch files

chenjiahan:
## Summary

This PR makes the Rslib CLI restart its watch process when `dev.watchFiles` entries use `type: 'reload-server'`, allowing tools such as Rstack to reload after `rstack.config.*` changes. Existing Rslib config and env file restart behavior is preserved.

## Checklist


Timeless0911 (approved):

chenjiahan:
I will improve Rsbuild's JS API to support this case

PR #1777, opened 2026-07-21
fix(core): avoid chunk filename conflicts in multi-compiler builds

Timeless0911:
## Summary

Multi-compiler builds can emit runtime and other initial chunks through `output.filename`, while async chunks use `output.chunkFilename`. Without a compiler-specific namespace, non-entry chunks can collide with chunks or entries emitted by another library config into the same directory.

- Preserve entry filenames while suffixing every non-entry initial/runtime and async chunk in multi-compiler builds with `~<index>`, starting from `~0`.
- Apply collision protection to the default filename and string `output.filename.js` templates. Filename functions remain user-owned, and explicit `tools.rspack.output.filename/chunkFilename` overrides remain authoritative.
- Extract the default ESM runtime only for bundleless or multi-entry builds, preserving the bundled single-entry single-file expectation.

| Build and `output.filename.js` | Entry | Runtime / other initial | Async |
| --- | --- | --- | --- |
| Single compiler, omitted or string | Keep the original template | Keep the original template | Infer from the original template |
| Multi-compiler, omitted or string (compiler `i`) | Keep the original template | Append `~i` | Infer from the template and append `~i` |
| Function | User-owned | User-owned | User-owned; Rslib does not infer or suffix `chunkFilename` |

## Related Links

- https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/rslib.config.ts#L82

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -362,7 +362,6 @@ const composeFormatConfig = ({
   umdName,
   pkgJson,
   enabledShims,
-  multiCompilerIndex,
   sourceEntry,
 }: {
   format: Format;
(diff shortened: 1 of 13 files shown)


Timeless0911:
@codex review

Timeless0911:
@codex review

JSerFeng (approved):

PR #266, opened 2024-10-10
fix: count in `optionalDependencies` in `autoExternal`.

fi3ework:
## Summary

- Count in https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies.
- No need to count https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bundledependencies since it's intended as a complement to `dependencies`.

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #1786, opened 2026-07-23
fix(dts): count `optionalDependencies` in autoExternal

elecmonkey:
## Summary

`optionalDependencies` was a typed and documented `autoExternal` option, but `calcBundledPackages` never added it to its dependency collection loop, so the option had no effect. This adds it to the loop, fixing `autoExternal: false` and `optionalDependencies: false` for optional dependencies.

## Related Links

Follow-up to #266.

## Checklist


elecmonkey:
@codex review

Timeless0911 (approved):

PR #1788, opened 2026-07-23
fix(dts): support `packageJson` and `exclude` in autoExternal

elecmonkey:
## Summary

Rslib passes Rsbuild’s `output.autoExternal` to `rsbuild-plugin-dts`. Rsbuild added the `packageJson` and `exclude` options, but the DTS plugin did not handle them.

This PR adds support for both options when calculating bundled packages.

## Checklist


elecmonkey:
@codex review

PR #1808, opened 2026-07-28
docs: remove Modern.js Module migration documentation in v1

elecmonkey:
## Summary

Thanks to the contributors who created and maintained this guide. The Modern.js Module migration is largely complete and has very few remaining users. This guide will not be part of the v1 documentation, but will remain available in the v0 documentation.


chenjiahan (approved):
LGTM

PR #1814, opened 2026-07-30
feat(core)!: enable createRequire parser by default

elecmonkey:
## Summary

Enable Rspack's `createRequire` parser by default so Rslib can statically analyze and bundle `createRequire()` calls.

The ESM and CJS outputs in the `tests/integration/shims/cjs` case remain correct and preserve the expected module semantics.

## Breaking Change

Static dependencies loaded through `createRequire()` are now bundled instead of being preserved for runtime resolution. Set `tools.rspack.module.parser.javascript.createRequire` to `false` to restore the previous behavior.

## Output Changes

For static `createRequire()` calls, the dependency is bundled and the call is replaced with `__webpack_require__()`.

In the shims integration case:

- ESM output bundles the CommonJS dependency and adds a shared Rspack runtime chunk.
- CJS output embeds the CommonJS module and scopes `"use strict"` to the ESM module factory instead of the entire chunk, preserving the original non-strict CommonJS semantics.
- Dynamic chunks and runtime behavior remain unchanged.

## Related Links

- https://github.com/web-infra-dev/rspack/pull/14783
- https://github.com/web-infra-dev/rspack/pull/14813

### DIFF
--- a/tests/integration/shims/index.test.ts
+++ b/tests/integration/shims/index.test.ts
@@ -146,11 +146,9 @@ describe('CJS shims', () => {
     expect(importMetaFilename).toBe(entryFiles.cjs);
 
     for (const code of [cjsCode, dynamicContent]) {
-      expect(
-        code.startsWith(
-          `"use strict";\nconst __rslib_import_meta_url__ = /*#__PURE__*/ function() {`,
-        ),
-      ).toBe(true);
+      expect(code).toContain(
+        'const __rslib_import_meta_url__ = /*#__PURE__*/ function() {',
+      );
     }
   });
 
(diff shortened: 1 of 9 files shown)


tests/integration/shims/index.test.ts
```diff
@@ -146,11 +146,9 @@ describe('CJS shims', () => {
     expect(importMetaFilename).toBe(entryFiles.cjs);
 
     for (const code of [cjsCode, dynamicContent]) {
-      expect(
-        code.startsWith(
-          `"use strict";\nconst __rslib_import_meta_url__ = /*#__PURE__*/ function() {`,
-        ),
-      ).toBe(true);
+      expect(code).toContain(
+        'const __rslib_import_meta_url__ = /*#__PURE__*/ function() {',
```
Timeless0911, line 150: help check if directive behave well now

elecmonkey: The behavior is correct. Once `./ok.cjs` is bundled, the chunk contains both strict ESM code and non-strict CommonJS code, so Rspack scopes `"use strict"` to the ESM module factory instead of the entire chunk. It remains the first statement in that factory and is effective for the ESM code, while the bundled CommonJS module keeps its original semantics.

elecmonkey: If we required the bundled file to also start with `"use strict"`, code that was originally non-strict would be forced into strict mode.

Timeless0911: LGTM

Timeless0911 (approved):

PR #1833, opened 2026-08-07
fix(core)!: Revert "feat(core)!: enable createRequire parser by default (#1814)"

elecmonkey:
Reverts #1814 and #1822, and removes the corresponding section from the v1 upgrade guide added in #1817.


tests/integration/shims/cjs/src/ok.cjs
```diff
@@ -1,6 +1 @@
-module.exports = {
-  value: 'ok',
-  isStrict: (function () {
-    return this === undefined;
-  })(),
-};
+module.exports = 'ok';
```
Timeless0911: we can keep the enhanced tests of strict mode in shims tests.

Timeless0911 (approved):

Timeless0911:
Add some tips about how to bundle `require()` in docs.

PR #1837, opened 2026-08-10
refactor(create-rslib)!: simplify Node.js templates

Timeless0911:
## Summary

Modern supported Node.js versions can synchronously require ESM, so maintaining separate pure ESM and dual ESM/CJS scaffolds adds unnecessary output and template choices.

- Consolidate the Node.js scaffolding into `node-js` and `node-ts`, publishing one ESM entry for both `import` and `require()`.
- Infer syntax compatibility from `engines.node`, preserve TypeScript declarations, and test the actual generated outputs.
- Update the English and Chinese documentation for the simplified templates and modern ESM interoperability.

The previous `node-esm-*` and `node-dual-*` template IDs are replaced by `node-*`.

## Checklist


packages/create-rslib/template-node-js/package.json
```diff
@@ -1,18 +1,11 @@
...
-  "name": "rslib-node-esm-js",
+  "name": "rslib-node-js",
   "version": "0.0.0",
   "type": "module",
-  "exports": {
-    ".": {
-      "import": "./dist/index.js"
-    }
-  },
+  "exports": "./dist/index.js",
```
chenjiahan, line 5: should we preserve the exports object?

Timeless0911: I think there is no need https://nodejs.org/api/packages.html#exports-sugar.

Timeless0911: Image

website/docs/en/guide/basic/output-format.mdx
```diff
@@ -33,7 +33,7 @@ For different shapes of libraries, the choice of module format may vary. Here ar
 
 #### ship [pure ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) package
 
-shipping only ESM is the best choice for libraries that are intended to be used in modern environments, such as browser applications or Node.js applications that support ESM. However, if the upstream library is in format of CJS, they only can import pure ESM by using dynamic import like `const pureEsmLib = await import('pure-esm-lib')`.
+Shipping only ESM is the best choice for libraries intended for modern environments, such as browser applications or Node.js applications that support ESM. [Node.js 20.19+ and 22.12+](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) can load ESM using `require()` when its module graph does not contain top-level `await`. CommonJS consumers on older Node.js versions, or those loading ESM with top-level `await`, should use dynamic import, such as `const pureEsmLib = await import('pure-esm-lib')`.
```
chenjiahan: It feels like a sentence is missing in the middle: “If consumers are still using CJS, …”

chenjiahan (dismissed):

PR #1841, opened 2026-08-11
fix(core): restore config key completions in defineConfig

chenjiahan:
## Summary

- Typing a partial key inside `defineConfig({ ... })` currently falls back to `Function.prototype` completions because TypeScript selects a callback overload.
- Use a const-generic definition with conditional return types while preserving Rslib top-level config validation, and cover object, sync, async, and dynamic definitions under Bundler and NodeNext resolution.

## Related Links

- https://github.com/web-infra-dev/rsbuild/pull/8248

## Checklist

### DIFF
--- a/packages/core/src/loadConfig.ts
+++ b/packages/core/src/loadConfig.ts
@@ -36,12 +36,22 @@ export type LoadConfigResult<Config = RslibConfig> = {
  * This function helps you to autocomplete configuration types.
  * It accepts a Rslib config object, or a function that returns a config.
  */
-export function defineConfig<const Config extends RslibConfig>(
-  config: (env: ConfigParams) => Config,
-): RslibConfigSyncFn;
-export function defineConfig<const Config extends RslibConfig>(
-  config: (env: ConfigParams) => Promise<Config>,
-): RslibConfigAsyncFn;
+export function defineConfig<
+  const Config extends RslibConfig,
+  const Definition extends
+    | Config
+    | ((env: ConfigParams) => Config)
+    | ((env: ConfigParams) => Promise<Config>),
+>(
+  config: Definition &
+    (Definition extends (...args: never[]) => unknown
+      ? unknown
+      : Record<Exclude<keyof Definition, keyof RslibConfig>, never>),
+): Definition extends (env: ConfigParams) => Promise<unknown>
+  ? RslibConfigAsyncFn
+  : Definition extends (env: ConfigParams) => unknown
+    ? RslibConfigSyncFn
+    : RslibConfig;
 export function defineConfig(config: RslibConfig): RslibConfig;
 export function defineConfig(
   config: RslibConfigDefinition,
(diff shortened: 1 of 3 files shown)


Timeless0911:
The config key completion issue is fixed, but the new signature introduces three type compatibility regressions:

1. Explicit generic callbacks no longer work:

```ts
defineConfig<RslibConfig>(() => ({
  lib: [],
}));
```

This previously compiled, but now reports `TS2743` because the first overload requires both `Config` and `Definition` type arguments.

2. Sync callbacks returning `any` or `never` are inferred as `RslibConfigAsyncFn`:

```ts
const config = defineConfig(() => JSON.parse(text));
// inferred as RslibConfigAsyncFn, although the callback is synchronous
```

This happens because the conditional return type checks the async signature first, and functions returning `any` or `never` are assignable to it.

3. Nested excess-property checking is lost:

```ts
defineConfig({
  lib: [{ typo: true }],
});
```

The previous object overload rejects `typo`, while the new generic overload accepts it. The current `Record<Exclude<...>, never>` guard only checks top-level keys and does not contextually validate nested config objects.

I suggest keeping the unified overload first, but changing it as follows:

```ts
export function defineConfig<
  const Config extends RslibConfig,
  const Definition extends
    | Config
    | ((env: ConfigParams) => Config)
    | ((env: ConfigParams) => Promise<Config>),
>(
  config: Definition &
    (Definition extends (...args: never[]) => unknown
      ? unknown
      : RslibConfig &
          Record<Exclude<keyof Definition, keyof RslibConfig>, never>),
): Definition extends RslibConfigSyncFn
  ? RslibConfigSyncFn
  : Definition extends RslibConfigAsyncFn
    ? RslibConfigAsyncFn
    : RslibConfig;

export function defineConfig<const Config extends RslibConfig>(
  config: (env: ConfigParams) => Config,
): RslibConfigSyncFn;

export function defineConfig<const Config extends RslibConfig>(
  config: (env: ConfigParams) => Promise<Config>,
): RslibConfigAsyncFn;
```

The important parts are:

- Intersect the object branch with `RslibConfig` to restore nested excess-property checking.
- Check `RslibConfigSyncFn` before `RslibConfigAsyncFn`, so sync callbacks returning `any` or `never` preserve the previous behavior.
- Restore the old callback overloads after the unified overload. This keeps single-generic calls compatible without bringing back the config completion issue, because normal inferred calls still use the first overload.

It would also be good to add type tests covering:

```ts
// Nested unknown keys must be rejected.
defineConfig({
  lib: [
    {
      // @ts-expect-error unknown config key
      typo: true,
    },
  ],
});

// Explicit generic callbacks must remain supported.
const explicitConfig: RslibConfigSyncFn = defineConfig<RslibConfig>(() => ({
  lib: [],
}));

// A sync callback returning any must remain sync.
const syncConfig: RslibConfigSyncFn = defineConfig(() => JSON.parse('{}'));
```

chenjiahan:
Nice catch. Fixed

PR #1853, opened 2026-08-17
fix(core): compile new URL module entries

### LINKED ISSUE #1825: [Feature]: resolve new URL as entry
### What problem does this feature solve?

After #1774, which closes #929, Rslib can compile a worker entry when `new URL()` is used directly inside a statically analyzable Web Worker constructor:

```ts
new Worker(new URL('./worker.ts', import.meta.url));
```

However, worker-pool libraries usually receive the worker module through an option instead of constructing a Web Worker directly.

For example, [Tinypool's documented API](https://github.com/tinylibs/tinypool#basic-usage) uses a `filename` option:

```ts
import Tinypool from 'tinypool';

const pool = new Tinypool({
  filename: new URL('./renderPageWorker.ts', import.meta.url).href,
});
```

The referenced source file is `src/renderPageWorker.ts`, but the URL is not recognized as a module entry.

According to the [Rslib static assets documentation](https://rslib.rs/guide/advanced/static-assets#new-url-imports), JavaScript and TypeScript files referenced by a plain `new URL()` are currently treated as URL assets. They bypass the built-in loaders and their original contents are emitted as assets.

As a result, the worker may be emitted as uncompiled TypeScript instead of an executable ESM module.

The current workaround is to manually declare every worker module as an additional entry:

```ts

export default defineConfig({
  source: {
    entry: {
      index: './src/index.ts',
      renderPageWorker: './src/renderPageWorker.ts',
    },
  },
});
```

The application code must then reference the expected output filename:

```ts
const pool = new Tinypool({
  filename: new URL('./renderPageWorker.js', import.meta.url).href,
});
```

This duplicates the worker relationship in the configuration, couples source code to the output layout, and becomes inconvenient when a library has multiple dynamically referenced entry modules.

The same requirement exists for Piscina, `child_process.fork()`, custom worker pools, and other APIs that accept the URL or filename of an executable module.

### What does the proposed API look like?

Rslib should treat a JavaScript or TypeScript module referenced by `new URL()` as a module entry and rewrite the URL to its compiled output.

For example:

```ts title="src/index.ts"
import Tinypool from 'tinypool';

export const pool = new Tinypool({
  filename: new URL('./renderPageWorker.ts', import.meta.url).href,
});
```

```ts title="src/renderPageWorker.ts"
export default async function renderPage(pathname: string): Promise<string> {
  return `<html><body>${pathname}</body></html>`;
}
```

Expected output:

```text
dist/
├── index.js
└── renderPageWorker.js
```

```js title="dist/index.js"
import Tinypool from 'tinypool';

const pool = new Tinypool({
  filename: new URL('./renderPageWorker.js', import.meta.url).href,
});

export { pool };
```

```js title="dist/renderPageWorker.js"
async function renderPage(pathname) {
  return `<html><body>${pathname}</body></html>`;
}

export { renderPage as default };
```

Expected semantics:

- Resolve the referenced module using Rslib's normal resolver.
- Compile the referenced file as an independent ESM entry instead of emitting its original source as a URL asset.
- Rewrite the `.ts` URL to the corresponding `.js` output file.
- In bundle mode, bundle the entry's dependencies into its output graph.
- In bundleless mode, compile the entry and preserve its module structure.
- Keep ordinary `new URL('./asset', import.meta.url)` behavior unchanged.
- Avoid adding a Webpack/Rspack runtime when the generated entry does not require one.

The important requirement is that statically analyzable JavaScript and TypeScript modules referenced by `new URL()` can be compiled as entries without also declaring them in `source.entry`.

kkkhs:
## Summary

Compile statically referenced JavaScript/TypeScript modules from `new URL(..., import.meta.url)` as ESM entries, and rewrite the URL to the generated JavaScript output.

## Related Links

Fixes #1825

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -585,12 +585,9 @@ const modifyRsbuildDefaultPlugin = ({
 
       // Part 2: configure URL parsing for library output.
       if (urlParserMode !== undefined) {
-        chain.module
-          .rule(NEW_URL_RULE)
-          .test(JS_EXTENSIONS_PATTERN)
-          .parser({
-            url: urlParserMode
-          });
+        chain.module.rule(NEW_URL_RULE).test(JS_EXTENSIONS_PATTERN).parser({
+          url: urlParserMode,
+        });
       }
 
       // Part 3: remove Rsbuild's `type: 'javascript/auto'` override.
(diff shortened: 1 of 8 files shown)


Timeless0911 (commented):
Thanks for the contribution and for taking the time to explore this use case. We appreciate the implementation and the test coverage added in this PR.

After reviewing the approach, we do not think this capability should be implemented in Rslib by scanning source text, injecting additional entries, and rewriting emitted assets. This happens outside Rspack's compilation model, so it cannot reliably handle resolver and loader semantics, transitive modules, watch and incremental builds, duplicate filenames, chunk/runtime selection, source maps, or asset metadata.

We plan to implement advanced `new URL(..., import.meta.url)` capabilities directly in Rspack. The design will use Rspack's URL parser and coordinate the relevant loader, module graph, chunking, and rendering stages so that resolution, compilation, output-path generation, caching, and watch mode work consistently.

This capability is still being designed. Thank you for your patience, and thanks again for the contribution—the use case and test fixture in this PR are valuable references for the future Rspack implementation.

PR #1854, opened 2026-08-18
refactor(create-rslib): layer template tsconfigs by scope

elecmonkey:
## Summary

Unify the TS template tsconfigs on one shape, modeled on `create-rstack`.

- Same option order and `/* modules */` / `/* type checking */` grouping in every
  root, filling in the missing `target`, `moduleDetection`,
  `allowImportingTsExtensions`, `noUnusedLocals` and `noUnusedParameters`.
- Every root is `include: ["src"]`; `template-svelte-ts` no longer pulls in
  `scripts` and `*.config.ts`.
- Every `tests/` gets a tsconfig extending the root with `rootDir: ".."` and
  `include: ["./"]` — `template-node-ts` and `template-svelte-ts` had none. Config
  and setup files are no longer covered by any tsconfig.
- Emit options are aligned across the two roots that emit declarations:
  `template-svelte-ts` gains the `outDir: "dist"` it was missing, and the test
  configs of both override `noEmit`.
- `types`: component roots stay browser-only (svelte's `["node"]` becomes
  `["@rslib/core/types", "svelte"]`), test configs add `node`. `@types/node` is
  added to the React, Solid and Vue templates, at the version the other templates
  already pin.
- Rspress no longer overwrites the generated root config:
  `template-rspress/react-ts/tsconfig.json` is removed and its `paths`
  self-reference moves to `docs/tsconfig.json`, with `src/index.ts` and the
  placeholder test updated.


packages/create-rslib/template-react-ts/tsconfig.json
```diff
@@ -2,13 +2,23 @@
   "compilerOptions": {
     "lib": ["DOM", "ES2022"],
     "jsx": "react-jsx",
+    "target": "ES2022",
+    "noEmit": true,
     "skipLibCheck": true,
+    "types": ["@rslib/core/types"],
```
Timeless0911, line 8: where is types node？

packages/create-rslib/template-solid-ts/tsconfig.json
```diff
@@ -3,13 +3,23 @@
     "lib": ["DOM", "ES2022"],
     "jsx": "preserve",
     "jsxImportSource": "solid-js",
+    "target": "ES2022",
+    "noEmit": true,
     "skipLibCheck": true,
+    "types": ["@rslib/core/types"],
```
Timeless0911, line 9: same

packages/create-rslib/template-svelte-ts/tsconfig.json
```diff
@@ -1,17 +1,26 @@
...
     "lib": ["DOM", "ES2022"],
-    "module": "preserve",
-    "moduleDetection": "force",
-    "moduleResolution": "bundler",
-    "resolveJsonModule": true,
-    "types": ["node"],
+    "target": "ES2022",
     "declaration": true,
     "emitDeclarationOnly": true,
+    "outDir": "dist",
```
Timeless0911, line 8: why we need outDir

elecmonkey: I think may write `outDir` alongside `"declaration": true` is more reasonable?

packages/create-rslib/template-react-ts/tsconfig.json
elecmonkey: It's component library. `src` may not include node types?

elecmonkey: node types devDeps is for test's tsconfig.json

Timeless0911 (approved):

PR #1858, opened 2026-08-19
chore(lint): enable Rstest lint rules

elecmonkey:
## Summary

- Upgrade `@rslint/core` from 0.8.0 to 0.8.1 and enable the recommended Rstest rules for test files.
- Address the most impactful rule, `rstest/no-conditional-expect`, by moving assertions out of conditional branches and using rejection assertions or early returns where appropriate. Keep the rules enabled because they exposed tests that could silently pass when an expected error was not thrown, demonstrating value beyond stylistic enforcement.
- Keep the intentional skipped test for an upstream issue with a targeted `// rslint-disable-next-line rstest/no-disabled-tests` comment.

## Related Links

- https://github.com/web-infra-dev/rslint/releases/tag/v0.8.1

## Checklist

### DIFF
--- a/rslint.config.ts
+++ b/rslint.config.ts
@@ -1,5 +1,4 @@
-import { defineConfig, js, rstestPlugin, ts } from '@rslint/core';
-import globals from 'globals';
+import { defineConfig, globals, js, rstestPlugin, ts } from '@rslint/core';
 
 export default defineConfig([
   {
--- a/rslint.config.ts
+++ b/rslint.config.ts
@@ -1,4 +1,4 @@
-import { defineConfig, js, ts } from '@rslint/core';
+import { defineConfig, js, rstestPlugin, ts } from '@rslint/core';
 import globals from 'globals';
 
 export default defineConfig([
@@ -43,6 +43,24 @@ export default defineConfig([
     files: ['**/*.cjs'],
     rules: {
       '@typescript-eslint/no-require-imports': 'off',
+    },
+  },
+  {
+    files: ['**/*.test.{ts,tsx}'],
+    ...rstestPlugin.configs.recommended,
+    rules: {
+      ...rstestPlugin.configs.recommended.rules,
+      'rstest/expect-expect': [
+        'warn',
+        {
+          assertFunctionNames: [
+            'expect',
+            'assert',
+            'expect*',
+            'createAndValidate',
+          ],
+        },
+      ],
     },
   },
 ]);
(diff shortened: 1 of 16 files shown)


tests/integration/config-check/index.test.ts
```diff
@@ -5,11 +5,13 @@ import { buildAndGetResults } from 'test-helper';
...
 test('should throw error when lib array not exists or empty', async () => {
   const fixturePath = join(__dirname, 'lib-array');
-  try {
-    await buildAndGetResults({ fixturePath });
-  } catch (error) {
-    expect(stripAnsi((error as Error).message)).toMatchInlineSnapshot(
-      `"Expect "lib" field to be a non-empty array, but got: []."`,
-    );
-  }
+  const error = await buildAndGetResults({ fixturePath }).then(
```
Timeless0911: can we update the test helper instead?

rslint.config.ts
```diff
@@ -45,4 +45,22 @@ export default defineConfig([
       '@typescript-eslint/no-require-imports': 'off',
     },
   },
+  {
+    files: ['**/*.test.{ts,tsx}'],
+    ...rstestPlugin.configs.recommended,
+    rules: {
```
Timeless0911, line 50: why we need to override this

tests/integration/config-check/index.test.ts
elecmonkey: Updated. Added `expectBuildToFail` to `test-helper` in 2ca6d57.

rslint.config.ts
elecmonkey: `rstest/expect-expect` cannot recognize custom assertion functions by default, causing false positives.

Timeless0911: any example in our repo？

elecmonkey: image

for example, in this case, no `expect` directly contains in the code block. `expect` are called in the `expectUseAdd`(it can not be static analyzed). so we need configure the function name it in this rule's option, the rule will consider `expectUseAdd` as assertion.

if we comment out these lines, we'll get: `Found 0 lint errors, 0 type errors and 22 warnings`

rslint.config.ts
```diff
@@ -1,4 +1,4 @@
-import { defineConfig, js, ts } from '@rslint/core';
+import { defineConfig, js, rstestPlugin, ts } from '@rslint/core';
 import globals from 'globals';
```
chenjiahan: We can remove the global dependency and import from Rslint

elecmonkey: oh yes. updated.

Timeless0911 (approved):

PR #1860, opened 2026-08-20
fix: keep scheme-less loader requests bundled

elecmonkey:
## Summary

- Keep requests without a URI scheme and containing `!` bundled, since Rspack treats `!` as a loader separator.
- Preserve externalization for scheme-based requests. For example, `data:text/javascript,export const bang = "a!b!c"` contains `!` as module content rather than loader syntax.

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -72,6 +72,17 @@ import {
 import { loadTsconfig } from './utils/tsconfig';
 import { composeWasmConfig, resolveWasmMode } from './wasm/compose';
 
+// Mirrors get_scheme() in rspack's loader runner. `builtin:` is intentionally
+// treated as scheme-less because Rspack uses it for inline builtin loaders.
+const hasScheme = (request: string) =>
+  /^[A-Za-z][A-Za-z0-9+-]*:/.test(request) &&
+  !/^builtin:/i.test(request) &&
+  !/^[A-Za-z]:[\\/#?]/.test(request);
+
+export function isInlineLoaderRequest(request: string): boolean {
+  return request.includes('!') && !hasScheme(request);
+}
+
 export function composeMinifyConfig(config: LibConfig): EnvironmentConfig {
   const minify = config.output?.minify;
   const format = config.format;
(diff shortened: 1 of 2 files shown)


Timeless0911:
@codex review

packages/core/src/config.ts
```diff
@@ -1283,6 +1283,15 @@ const composeBundlelessExternalConfig = (
     request.includes('rspack-vue-loader') &&
     (suffix ? request.includes(suffix) : true);
 
+  // Mirrors get_scheme() in rspack's loader runner. Requests without a URI
+  // scheme are split on every `!` by NormalModuleFactory, so any `!` in such
+  // a request identifies an inline loader request and must not be externalized.
+  const hasScheme = (request: string) =>
+    /^[A-Za-z][A-Za-z0-9+-]*:/.test(request) &&
```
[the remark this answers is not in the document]
elecmonkey: This does not apply here because Rspack’s current get_scheme parser also rejects `.` in scheme names, so `foo.bar:payload!text` is correctly treated as an inline loader request.

Timeless0911:
## Issue

`hasScheme` incorrectly treats `builtin:` as a URI scheme. Rspack treats `builtin:swc-loader!./value.js` as an inline loader request, so it may leak into bundleless output.

## Missing Tests

This PR currently adds no tests for the new detection logic. Please add unit tests covering key cases.

elecmonkey:
> ## Issue
> `hasScheme` incorrectly treats `builtin:` as a URI scheme. Rspack treats `builtin:swc-loader!./value.js` as an inline loader request, so it may leak into bundleless output.
> 
> ## Missing Tests
> This PR currently adds no tests for the new detection logic. Please add unit tests covering key cases.

fixed in [e4e9ea5](https://github.com/web-infra-dev/rslib/pull/1860/commits/e4e9ea5a51789efd8780848d09e556926a8ab25d)

packages/core/src/config.ts
```diff
@@ -72,6 +72,17 @@ import {
 import { loadTsconfig } from './utils/tsconfig';
 import { composeWasmConfig, resolveWasmMode } from './wasm/compose';
 
+// Mirrors get_scheme() in rspack's loader runner. `builtin:` is intentionally
```
Timeless0911: comment a link which redirects to rspack code here

packages/core/src/config.ts
```diff
@@ -72,6 +72,17 @@ import {
...
+// treated as scheme-less because Rspack uses it for inline builtin loaders.
+const hasScheme = (request: string) =>
+  /^[A-Za-z][A-Za-z0-9+-]*:/.test(request) &&
+  !/^builtin:/i.test(request) &&
```
Timeless0911: `builtin`   should this be case insensitive

elecmonkey: /i already makes the regex case-insensitive. Added some tests.

Timeless0911 (dismissed):

Timeless0911 (approved):

PR #1861, opened 2026-08-20
feat(create-rslib): use bundleless mode for Svelte templates

elecmonkey:
## Summary

This PR makes the Svelte JS and TS library templates bundleless by default and adds integration coverage for bundled and bundleless ESM output, including generated CSS and scoped styles. Snapshots are refreshed against the latest main to validate the corrected bundleless CSS imports.

## Checklist


tests/integration/svelte/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "svelte-test",
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "devDependencies": {
+    "@rsbuild/plugin-less": "catalog:",
+    "@rsbuild/plugin-svelte": "catalog:",
+    "svelte": "catalog:"
+  },
+  "peerDependencies": {
```
Timeless0911: peer is no need for test

tests/integration/svelte/rslib.config.ts
```diff
@@ -0,0 +1,27 @@
...
+      output: {
+        target: 'web',
+        distPath: 'dist/bundleless',
+      },
+    }),
+    // bundle, ESM
+    generateBundleEsmConfig({
+      plugins: [pluginSvelte()],
+      output: {
+        target: 'web',
```
Timeless0911: set to global

pnpm-workspace.yaml
```diff
@@ -121,9 +123,11 @@ catalogs:
     'typescript6-api': 'npm:@typescript/typescript6@^6.0.2'
 
 allowBuilds:
```
Timeless0911, line 125: should be false

packages/create-rslib/template-svelte-js/rslib.config.js
```diff
@@ -2,6 +2,7 @@ import { pluginSvelte } from '@rsbuild/plugin-svelte';
 import { defineConfig } from '@rslib/core';
 
 export default defineConfig({
+  bundle: false,
```
Timeless0911, line 5: we can remove the css export from package.json

tests/integration/svelte/package.json
elecmonkey: removing the peer dep would cause svelte runtime included in the snapshot in bundle mode. so let’s only test bundleless mode.

Timeless0911: ok, bundleless mode is enough.

packages/create-rslib/template-svelte-js/package.json
```diff
@@ -3,8 +3,7 @@
   "version": "0.0.0",
   "type": "module",
   "exports": {
-    ".": "./dist/index.js",
-    "./style.css": "./dist/index.css"
+    ".": "./dist/index.js"
```
Timeless0911: no need to use object

elecmonkey: Updated.

Timeless0911 (dismissed):

packages/create-rslib/template-svelte-js/package.json
Timeless0911: test failed

Timeless0911 (dismissed):

Timeless0911 (approved):

PR #1872, opened 2026-08-24
fix(core): avoid extracting unused Node archive files

### LINKED ISSUE #1871: [Bug]: multi-target executable build requires Windows symlink privileges
### Version

`@rslib/core@1.0.0-rc.1` / commit `af41f2ed7523e2c5130084d80cc058994a628049`

### Details

A multi-target `experiments.exe` build on Windows extracts the complete Linux and macOS Node.js archives with `tar`. Those archives contain `bin/npm` and `bin/npx` symlinks. On a normal Windows account without symlink privileges, BSD tar fails while creating those entries, so Rslib never reaches the Node executable that it actually needs.

The existing Windows CI job succeeds because its runner can create the archive symlinks, but the same checked-in integration test fails under a non-elevated local account.

### Reproduce link

https://github.com/web-infra-dev/rslib/tree/af41f2ed7523e2c5130084d80cc058994a628049/tests/integration/exe

### Reproduce steps

On Windows, use Node 25.9.0 from a non-elevated account without symlink privileges, then run:

```powershell
pnpm install --frozen-lockfile
pnpm run build
cd tests
pnpm exec rstest run --project integration-exe
```

The `should generate multi-target executables` test fails while extracting `node-v25.9.0-linux-x64.tar.gz` (and then the macOS archive):

```text
bin/npx: Can't create ...
bin/npm: Can't create ...
tar: Error exit delayed from previous errors.
```

### Expected behavior

Cross-platform targets should be generated without requiring Windows symlink privileges. Rslib only consumes `<archive-root>/bin/node`, so extracting that member instead of the whole tarball avoids the unrelated symlinks and reduces extraction work.

### Environment

- Windows 11
- Node.js 25.9.0
- pnpm 11.22.0
- non-elevated account without symlink privileges

dajiaohuang:
## Summary

- extract only the Node executable from Unix target archives
- avoid unrelated `npm` and `npx` symlink creation on Windows
- reduce cross-target archive extraction work

## Related Links

Fixes #1871

## Checklist

## Validation

- `pnpm --filter @rslib/core run build`
- `pnpm exec rslint packages/core/src/exe/download.ts`
- `pnpm dlx node@25.9.0 ..\node_modules\@rstest\core\bin\rstest.js run --project integration-exe` from `tests` (5/5 passed)
- exact single-member `tar` extraction on Windows without symlink privileges (1 file extracted)

### DIFF
--- a/packages/core/src/exe/download.ts
+++ b/packages/core/src/exe/download.ts
@@ -240,6 +240,7 @@ const extractArchive = async ({
     archivePath,
     '-C',
     path.dirname(extractDir),
+    `${path.basename(extractDir)}/bin/node`,
   ]);
 };
 


Timeless0911 (approved):
LGTM

Issue #1291, opened 2025-10-24
[Feature]: Add an option to automatically generate exports fields in package.json


Karibash:
### What problem does this feature solve?

Currently, `rslib` aims for a zero-configuration experience by treating `package.json` as the single source of truth (see [#62](https://github.com/web-infra-dev/rslib/issues/62)).
However, since `rslib` does not exclude `*.test.ts` files by default, it’s difficult to achieve a truly configuration-free setup.

To improve this, it would be helpful to introduce an option that automatically generates the following fields in `package.json` based on the build entries:
- main
- module
- types
- exports

This would make the build output and package metadata consistent without requiring manual edits to `package.json`.

For reference, tools like `tsdown` [already provide a similar feature](https://tsdown.dev/options/package-exports) that derives these fields from the build entry configuration.
Implementing something comparable in `rslib` would simplify the developer experience and reduce maintenance overhead.

### What does the proposed API look like?

```ts
import { defineConfig } from '@rslib/core'

export default defineConfig({
  output: {
    exports: true,
  },
})
```

When enabled, `rslib` automatically generates `exports`, `main`, `module`, and `types` fields in package.json based on the build entries.

Timeless0911:
I've been paying attention to this feature before. I think this is a good way to help library developers reduce their burden when developing simple npm packages, but the limitation is that its accuracy is difficult to guarantee thus we should review the generated fields before publishing. And when users need to modify or override some small parts of generated exports, modifying it through function way seems not direct.

I prefer to introduce this experimental feature through a plugin maybe named `rsbuild-plugin-packagejson-exports`, e.g. [rsbuild-plugin-publint](https://github.com/rspack-contrib/rsbuild-plugin-publint) and [rsbuild-plugin-arethetypeswrong
](https://github.com/rspack-contrib/rsbuild-plugin-arethetypeswrong).

Timeless0911:
> I prefer to introduce this experimental feature through a plugin maybe named rsbuild-plugin-packagejson-exports, e.g. [rsbuild-plugin-publint](https://github.com/rspack-contrib/rsbuild-plugin-publint) and [rsbuild-plugin-arethetypeswrong](https://github.com/rspack-contrib/rsbuild-plugin-arethetypeswrong).

Considering that this feature should not require additional dependencies, it seems better to have it as a built-in configuration. We will add this feature in the near future, and if you are interested, welcome to contribute a pull request.

PR #1903, opened 2026-09-04
feat(core): add output.exports option

MissGwen:
## Summary

This PR adds an `output.exports` option to generate the `exports` field in `package.json` from build entries.

- Supports ESM and CJS outputs, including `import` and `require` conditions.
- Supports single and multiple entries.
- Skips generation in watch mode.

This is the first iteration and only covers `exports`. If the remaining fields proposed in the issue are still needed, I’d be happy to add `main`, `module`, and `types` in a follow-up PR.

## Related links

- https://github.com/web-infra-dev/rslib/issues/1291

## Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1859,6 +1859,7 @@ export async function composeCreateRsbuildConfig(
   rslibConfig: RslibConfig,
 ): Promise<RsbuildConfigWithLibInfo[]> {
   const constantRsbuildConfig = await createConstantRsbuildConfig();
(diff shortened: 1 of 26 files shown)


elecmonkey:
Thank you for your contribution! The team is still discussing and thinking through the implementation approach. We’ll share any updates or continue the discussion here once we make progress!

MissGwen:
> Thank you for your contribution! The team is still discussing and thinking through the implementation approach. We’ll share any updates or continue the discussion here once we make progress!

Thanks for the update! Once the approach is settled, I’d be happy to update this PR accordingly and help with the remaining work in #1291 if needed.

PR #1909, opened 2026-09-07
feat(wasm): support inlining wasm binaries with the `?inline` query

elecmonkey:
## Summary

Support importing small WebAssembly binaries with `?inline` to embed them directly into JavaScript.

```js
import { add } from './add.wasm?inline';

export const sum = (a, b) => add(a, b);
```

The binary is instantiated with top-level `await` and exposes the exports declared by the WebAssembly module as ESM exports. This requires ESM output.

In bundleless builds, the binary is embedded into each importing JavaScript file. WebAssembly imports are resolved as module dependencies.


huanglinfei091-cmd:
Hi @elecmonkey — I noticed the `ut (ubuntu-latest, 24) / test` job failed at `Unit Test` in [this GitHub Actions run](https://github.com/web-infra-dev/rslib/actions/runs/34075715726) for this PR.

I’m testing [RunBack](https://github.com/huanglinfei091-cmd/runback), an early alpha CLI that reconstructs a failed public Actions run locally with `act`. For this run, the useful question is whether the Node 24 unit-test shard produces the same assertion, or whether reconstruction stops with an honest blocked result.

```bash
runback doctor
runback https://github.com/web-infra-dev/rslib/actions/runs/34075715726
```

The Linux amd64 binary and checksum are in the [v0.1.0-alpha release](https://github.com/huanglinfei091-cmd/runback/releases/tag/v0.1.0-alpha). If you try it, the verdict (`SAME_FAILURE`, `DIFFERENT_FAILURE`, or `REPLAY_BLOCKED`), Stage/Cause, and any install friction would be useful. Please do not post tokens or secrets. I’m asking for candid alpha evidence rather than a star or promotion; a failed reproduction is useful too.

packages/core/src/wasm/inlineLoader.ts
```diff
@@ -0,0 +1,29 @@
...
+  generateWasmInlineModule,
+  WASM_INLINE_ISSUER_QUERY,
+} from './inline.js';
+
+function wasmInlineLoader(
+  this: Rspack.LoaderContext<{ format: Format }>,
+  content: Buffer,
+): string {
+  const { format } = this.getOptions();
+  if (format !== 'esm') {
```
Timeless0911, line 13: I think this should be a breaking change. Why not register inline loader only in esm format?

packages/core/src/wasm/compose.ts
```diff
@@ -52,38 +71,57 @@ export const composeWasmConfig = ({
...
+          'rslib-wasm-inline',
+          (factory) => {
+            factory.hooks.resolve.tap('rslib-wasm-inline', (data) => {
+              if (isWasmInlineRequest(data.request)) {
+                // Modern-module output extracts shared modules, so scope inline modules to their importer.
+                const issuer =
+                  new URLSearchParams(
+                    data.request.split('?')[1]?.split('#')[0],
+                  ).get(WASM_INLINE_ISSUER_QUERY) ?? data.contextInfo.issuer;
+                data.request = data.request.replace(
```
Timeless0911, line 91: Could we avoid appending the issuer parameter when it is already present? If an already-mutated request ever re-enters this hook, this would add duplicate `__rslib_wasm_inline_issuer` parameters. A small `searchParams.has(WASM_INLINE_ISSUER_QUERY)` guard would make the transformation idempotent.

packages/core/src/wasm/compose.ts
```diff
@@ -52,38 +71,57 @@ export const composeWasmConfig = ({
...
+              externals,
+            },
+          }
+        : {},
     config: {
       tools: {
-        rspack: {
-          plugins: [new WasmPreservePlugin(preserveOptions.outBase)],
-        },
+        bundlerChain: applyWasmInlineRule(format),
```
Timeless0911, line 123: Since this rule is registered regardless of `mode`, `?inline` imports remain enabled when `wasm: false` (in ESM format), while plain `.wasm` imports are left untouched. Is this per-import opt-in behavior intentional? If so, could we document it to clarify how `?inline` interacts with the global `wasm` option?

Timeless0911:
no documents about inline feature in wasm chapter?

packages/core/src/wasm/compose.ts
elecmonkey: I think `wasm: false` should also prevent processing `?inline`. This is an oversight.

packages/core/src/wasm/inlineLoader.ts
elecmonkey: Non-ESM `.wasm?inline` can't produce usable output today, so I don't think this is a breaking change; it turns a silently broken build into a clear error instead.

packages/core/src/wasm/compose.ts
elecmonkey: Thinking about this again, I don't think handling an already-present issuer query here is meaningful, so I dropped the read instead of adding the guard.

> The only way a request reaches this hook with the query already on it is a wasm whose import section names another `.wasm?inline` module, which isn't standard under the wasm ESM integration. And the read was only tolerating that halfway — it picked up the existing value but still appended a second one. Removing it makes the assumption explicit: requests coming in here never carry the query.

packages/core/src/wasm/compose.ts
elecmonkey: Fixed.

packages/core/src/wasm/inlineLoader.ts
```diff
@@ -0,0 +1,29 @@
...
+    throw new Error(
+      `Importing wasm with the "?inline" query only supports the "esm" format, but the current format is "${format}".`,
```
Timeless0911, line 15:
```suggestion
      `WASM imports with "?inline" require the "esm" format, but the current format is "${format}".`,
```

website/docs/zh/guide/advanced/wasm.mdx
```diff
@@ -309,3 +322,19 @@ TypeScript 未内置 `.wasm` 文件的模块声明。使用 TypeScript 时，你
 ```ts title="src/add.d.wasm.ts"
 export function add(a: number, b: number): number;
 ```
+
+### 内联导入 \{#inline-imports}
+
+使用 TypeScript 时，在 ambient module 中声明每个 `?inline` WASM 模块的导出类型：
```
Timeless0911: `@rslib/core/types` includes `@rsbuild/core/types`, which declares `*?inline` as a string module. With the standard Rslib TypeScript setup, this declaration takes precedence over the suggested WASM declaration, so the default export is typed as string and named exports are unavailable. Could we add a type test using `@rslib/core/types` and resolve this conflict before documenting the workaround?

elecmonkey: Rsbuild declares `*?inline` as a string, and Rslib users receive this declaration through `@rslib/core/types`. The previous docs used `*add.wasm?inline`, but TypeScript prioritizes wildcard patterns by the prefix before `*`. Since both prefixes are empty, the result depends on declaration order.

I found that adding `export {}` makes the dts file a module, allowing `declare module './add.wasm?inline'` to augment the exact import instead.

elecmonkey:
I suggest merging the current version, since `.d.mts` doesn't act same as `export {}`. TypeScript's extension-based module detection intentionally excludes dts files. Also, I verified that the configuration described in the current docs produces a type error without `@rslib/core/types`, because it is an augmentation for an existing, broader `*?inline` declaration provided by Rsbuild.

Timeless0911:
@codex review

Timeless0911 (approved):

